PR #7, opened 2019-02-20
Limit Agent API calls per second to NSXv3

tsimchev:
The limit of NSX-T is 100 calls per second

Control Parameters
sync_pool_size := Number of threads synchronizing OpenStack and NSXv3.
sync_requests_per_second := Requests per second synchronizing OpenStack and NSXv3.
nsxv3_requests_per_second := Requests per second to NSXv3 Manager. NSXv3 limit 100 req/s


notandy:
hey thanks!, would you mind to add ratelimiter to requirements.txt, so our pip install procedure considers it automatically? Thanks

PR #23, opened 2019-08-16
Development/queens migration fixes

tsimchev:


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -157,9 +156,9 @@ def get_active_workers(self):
     def sync(self):
         with LockManager.get_lock(AGENT_SYNCHRONIZATION_LOCK):
             msg = "FULL SYNCHRONIZATION CYCLE - {}"
-            max_calls = cfg.CONF.AGENT.sync_requests_per_second
+            rate = cfg.CONF.AGENT.sync_requests_per_second
 
-            @RateLimiter(max_calls=max_calls, period=1)
+            @Scheduler(rate=rate, limit=1)
```
joker-official: Have you thought about making the `Scheduler` a singleton? Because this is another instance of the `Scheduler` than then `api_scheduler` down below and thus they won't interact and the limit could still be exceeded.

tsimchev: The low-level Scheduler is related to the API calls where the upper-level one is related only to the number of OpenStack objects about to be synced.
My idea is to provide a way to control the number of sync objects, where the api_scheduler will handle both the objects coming from the sync and this coming from the regular operation. That is the reason I make two instances.

joker-official: Ah, ok. Thank you for the explanation.

PR #25, opened 2019-10-01
Make Neutron RPC calls asynchronous

tsimchev:
Agent receives updates from Neutron or Scheduled synchronization,
now all the calls will be placed in a single priority queue,
where the RPC calls will be with highest priority,
and the Scheduled with lower.


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -124,99 +192,86 @@ def security_group_delete(self, security_group_id):
...
+
+    def _sync_inventory_shallow(self):
+        LOG.info("Shallow synchronization started")
+        sg_query = self.rpc.get_security_group_revision_tuples
+        qos_query = self.rpc.get_qos_policy_revision_tuples
+        port_query = self.rpc.get_port_revision_tuples
+
+        outdated_ips, orphaned_ips = self._sync_get_content(
+            sdk_model=IPSet(), os_query=sg_query)
+        outdated_qos, orphaned_qos = self._sync_get_content(
```
joker-official: Why is this done before the task using it is started? With it having the lowest priority, the information might be out-of-date for a long time.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -124,99 +192,86 @@ def security_group_delete(self, security_group_id):
...
+        self.synchronizer.submit_task(
+            1, outdated_ips, self.security_group_member_updated)
+        self.synchronizer.submit_task(
+            2, outdated_ips, self.security_group_rule_updated)
+        self.synchronizer.submit_task(
+            3, orphaned_ips, self.sync_security_group_orphaned)
+        self.synchronizer.submit_task(
+            4, outdated_lps, self.sync_port)
+        self.synchronizer.submit_task(
+            5, outdated_qos, self.sync_qos)
```
joker-official: How do you make sure that task is run at all with many high-priority tasks coming in all the time, i.e. how do you prevent starving and thus becoming outdated? One option I've see in other software was having a different number of workers assigned per priority thus making high-priority tasks more likely to schedule.

tsimchev: Synchronization is scheduled for ID, so the update will always comply with the latest state, though it could be unnecessary at the point of its application.

I believe that real-time events has to be handled first and only then the sync tasks can be handled. The real bottleneck is the NSX-T API, thus if the agent cannot handle real-time tasks, then we have a problem and the agent synchronization cycle can detect that as it will not be able to complete. 
If you prefer to use two queues and want to have reservation I am fine to apply this change.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
tsimchev: agree, will apply the change

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -65,56 +136,53 @@ def security_group_member_updated(self, security_group_id):
...
+                fwr["remote_group_id"] = revs_ips[remote_group_id]
 
-                if name in revs_fwr:
-                    # If the rule is disabled recreate it
-                    if not meta_fwr.get(name).get("FirewallRule.disabled"):
-                        del revs_fwr[name]
-                        continue
+            if name in revs_fwr:
+                # If the rule is disabled recreate it
+                if not meta_fwr.get(name).get("FirewallRule.disabled"):
```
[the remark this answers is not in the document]
tsimchev: There no reason, tabulation mistake.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -46,15 +47,85 @@ def is_migration_enabled():
...
+
+    def __init__(self, context, agent, sg_agent, nsxv3, vsphere, rpc):
+        super(NSXv3AgentManagerRpcCallBackBase, self).__init__(
+            context, agent, sg_agent)
+        self.nsxv3 = nsxv3
+        self.vsphere = vsphere
+        self.rpc = rpc
+
+        scheduler = Scheduler(
+            rate=cfg.CONF.AGENT.sync_requests_per_second, limit=1)
```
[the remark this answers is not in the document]
tsimchev: will apply a fix.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
joker-official: With outdated I mean the data in the manager that this task is supposed to sync.

With "detect that" you mean we could look into the logs? Bug couldn't we also look into the logs for the log-messages of the `Scheduler` telling us that we got rate-limited? Do we even get those for the syncprocessor's scheduler?

I don't prefer anything currently. I'd like to understand the pros and cons. What's the worst case if that happens if we cannot run the `sync_qos` but all the other `sync_*` run? That's all just to make sure the data is actually correct, right? It should have already been updated in those other notification handlers?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -124,99 +192,86 @@ def security_group_delete(self, security_group_id):
     def security_groups_member_updated(self, context, **kwargs):
         o = kwargs["security_groups"]
         o = o if isinstance(o, list) else [o]
-        for id in o:
-            self.security_group_member_updated(id)
+        self.synchronizer.submit_task(0, o,
```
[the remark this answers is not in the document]
tsimchev: agree

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
tsimchev: Yes, with "detect that" I mean looking into the logs.
Rate limiter can help from preventing to push more objects in the Queue per second as we would like, but currently, the actual handling depends on the number of run-time requests and the capacity of the consumer.
There could be a situation where sync rate limiter is not hit at all but non of the objects for synchronization can processed.

I think we need to do some whiteboarding and define how to address this.

PR #43, opened 2020-10-02
Configure default revision_number for IPSet

tsimchev:
Issue
Unable to populate rules referencing to a remote security group (IPSet)
that does not have reviion number

Resolution
Set default revision_number to all IPSets during creation

Side Effect
For existing security groups where this issue is still active
force update with default revision_number.
The update should be realized during the firt shallow or full sync.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_facada.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_facada.py
@@ -614,15 +614,10 @@ class NSXv3Facada(nsxv3_client.NSXv3ClientImpl):
         default_tag = Tag(scope=nsxv3_constants.NSXV3_REVISION_SCOPE, tag="-1")
 
         sdk_model.tags = sdk_model.tags if sdk_model.tags else []
-        revision = None
-        for tag in sdk_model.tags:
-            if tag.scope == default_tag.scope:
-                revision = tag.tag
-                break
-        if revision is None or not sdk_model.tags:
-            sdk_model.tags.append(default_tag)
-            sdk_model = self.update(sdk_service=IpSets, sdk_model=sdk_model)
-        return sdk_model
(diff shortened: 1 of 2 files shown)


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_facada.py
```diff
@@ -612,6 +607,29 @@ def update_security_group_capabilities(self,
...
+        """
+        default_tag = Tag(scope=nsxv3_constants.NSXV3_REVISION_SCOPE, tag="-1")
+
+        sdk_model.tags = sdk_model.tags if sdk_model.tags else []
+        revision = None
+        for tag in sdk_model.tags:
+            if tag.scope == default_tag.scope:
+                revision = tag.tag
+                break
+        if revision is None or not sdk_model.tags:
```
joker-official, line 622: I don't think the condition after the `or` makes sense with above's loop. Revision will always be `None` if there are no tags. So I'd remove it to have less code to confuse people.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_facada.py
```diff
@@ -612,6 +607,29 @@ def update_security_group_capabilities(self,
...
+            sdk_model.tags.append(default_tag)
+            sdk_model = self.update(sdk_service=IpSets, sdk_model=sdk_model)
+        return sdk_model
```
joker-official, line 625: I'll write that differently, but that might be taste:
```python
if default_tag.scope in (t.scope for t in sdk_model.tags):
    return sdk_model
sdk_model.append(default_tag)
return self.update(sdk_service=IpSets, sdk_model=sdk_model)
```

joker-official (approved):
Two minor code improvements. I think it should work.

joker-official (approved):

PR #44, opened 2020-10-04
Development/migrate to segments

tsimchev:


joroaf (changes requested):
I think this is outdated and needs to be canceled

PR #49, opened 2021-01-18
Raise ObjectNotFound as expected exception on server

joker-official:
This should lead to the exception not being reported on neutron-server
as an actual error. Instead, the exceptions are just packed up and sent
to the client, where they are raised again.


joker-official:
Haven't tested this, but it looks like Nova works like this. This should remove a lot of sentry entries (we usually have them doubled - once reported by neutron-server and once by the agent, afaics).

notandy (approved):

PR #51, opened 2021-02-18
Automatic switch between Management and Policy APIs

tsimchev:
Here is the flow of API Handlers
if version < 2.5
    API Handlers
    events - Management
    synchronization shallow - Management
    synchronization full - Management
if version >= 2.5 and policy tag not in TransportZone.tags
    events - Policy
        deletion of non-existing policy API will delete the Management object
    synchronization shallow - disable
    synchronization full - disable
    synchronization migration to policy API - enable
        on completion add policy tag to the transportZone.tags
    synchronization shallow - enable
    synchronization full - enable
else
    events - Policy
    synchronization shallow - Policy
    synchronization full - Policy

### DIFF
--- a/networking_nsxv3/api/rpc.py
+++ b/networking_nsxv3/api/rpc.py
@@ -8,8 +8,8 @@ from osprofiler.profiler import trace_cls
 
 from networking_nsxv3.common import constants as nsxv3_constants
 from networking_nsxv3.db import db
-from networking_nsxv3.common import exceptions
 
+from neutron_lib import exceptions
 from neutron_lib.plugins import directory
 from oslo_log import helpers as log_helpers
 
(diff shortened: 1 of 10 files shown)


notandy:
Hi, I tested the PR a bit:

For me, it breaks a lot. First thing, the `sync_security_groups_monotonically` function blocks continuously due to the usage of the sync_scheduler and completion-wait.
This doesn't work out, it completely blocks the sync_inventory loop, exhausting all threads and also blocks the agent from running _report_state which therefore causes an endless restart-loop (We have a liveness-check that works on the status of the agent according to neutron). 

I would recommend to get rid of a "migration" sync function and move the policy realization-check into the (shadow-)sync which limit's the total amount of task in the queue.

tsimchev:
@notandy the main reason for the blocking loop is that I am waiting for all security group in the queue to get realized before continue and this is the problem, some of the groups does not get realized  forever for some reason.

A possible solution would be to have timeout for security groups updates older than certain age and log errors for these.

notandy:
closing becase migration is already implemented

PR #52, opened 2021-02-19
Pass object via id to ObjectNotFound

sebageek:
ObjectNotFound does not take any positional arguments, but takes an id
as keyword argument. Strictly speaking we don't have an id, but the
exception would look decent enough:

"Object Security Group '7bf...' not found."

### DIFF
--- a/networking_nsxv3/db/db.py
+++ b/networking_nsxv3/db/db.py
@@ -21,11 +21,10 @@ from neutron_lib import exceptions
 
 
 def _validate_one(result, error):
-    msg = "{} not found in Neutron."
     if result:
         return result
     else:
-        raise exceptions.ObjectNotFound(msg.format(error))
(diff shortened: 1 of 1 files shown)


networking_nsxv3/db/db.py
```diff
@@ -25,7 +25,7 @@ def _validate_one(result, error):
     if result:
         return result
     else:
-        raise exceptions.ObjectNotFound(msg.format(error))
+        raise exceptions.ObjectNotFound(id=msg.format(error))
```
joker-official, line 28: I think you want to just pass `error` here instead of `msg.format(error)`. Currently it looks like `Object 'FirewallSection=7a7sfasdf not found in Neutron' not found.`.

joker-official (changes requested):

joker-official (approved):

PR #54, opened 2021-03-11
Ignore non nsxt trunk ports

sebageek:


networking_nsxv3/services/trunk/drivers/nsxv3/trunk.py
```diff
@@ -57,24 +57,29 @@ def register(self, resource, event, trigger, payload=None):
...
 
-    def trunk_update(self, resource, event, trunk_plugin, payload):
-        LOG.info("Trunk %s update called", payload.trunk_id)
-
     def trunk_delete(self, resource, event, trunk_plugin, payload):
+        ctx = context.get_admin_context()
+        parent = self.core_plugin.get_port(ctx, payload.original_trunk.port_id)
+        if not self.is_interface_compatible(parent[portbindings.VIF_TYPE]):
+            return
+
```
notandy, line 82: same here

networking_nsxv3/services/trunk/drivers/nsxv3/trunk.py
```diff
@@ -57,24 +57,29 @@ def register(self, resource, event, trigger, payload=None):
...
-        LOG.info("NSXv3 trunk driver initialized.")
 
+        LOG.info("NSXv3 trunk driver initialized.")
 
     def trunk_create(self, resource, event, trunk_plugin, payload):
+        ctx = context.get_admin_context()
+        parent = self.core_plugin.get_port(ctx, payload.current_trunk.port_id)
+        if not self.is_interface_compatible(parent[portbindings.VIF_TYPE]):
+            return
+
```
notandy, line 71: `_bind_subports` acquires the admin_context and does the same check, duplicate code?

`_bind_subports` could alternativley return a boolean if it accepts the port.

notandy (changes requested):

networking_nsxv3/services/trunk/drivers/nsxv3/trunk.py
sebageek: `_bind_subports()` need the parent port anyways, so for all subport related operations I do the check in the `_bind_subports()` method. `trunk_create()` / `trunk_delete()` call `_bind_subports()` but I find it somewhat unintuitive that this method returns a boolean, deciding if this is a port we need to act upon. In the ACI driver I have very similar code (@tsimchev adapted my original version for nsx-t). There I tried to refactor this check into an extra method, which was just more clunky to handle for me. Having this check only in `_bind_subports()` will also not work at the point where we add precommit-hooks. 

I'd be in favor of keeping the presented version, but if my arguments were not convincing enough I would be open to let `_bind_subports()` return a bool and revisit this when the development of the trunk extension has advanced further.

notandy: `get_admin_context` creates a brand new session on every invocation, that's the reason such context's cached and passed to all sub-procedures in upstream. Returining boolean seemed to me very straigh forward, but choose your weapons. 

I am also not understanding what do you mean by adding precommit-hooks, are they needed? I thought the nsx-t trunk implementation is already complete.

sebageek: Hmm, you're right, as I'm creating the ctx each time anyway I can just pass that to `_bind_subports()`. New version attached. ;)

notandy:
looks good, if you don't mind, I would merge it later since I need to roll-out nsx-t virtual-switch delay and auto-migration code first.

tsimchev-vmw:
@notandy just synced with @sebageek to understand `is_interface_compatible` behaviour, I am good with the changes

PR #61, opened 2021-04-02
Clean up orphan IPSets every time when agent boots

tsimchev-vmw:
Find all orphan IPSets which used to be one-to-one associated with a Rules from Management API
These IPSets share the same name as the rule they have been associated with
The identification is implemented by querying all IPSets IDs of type Rule CIDR and
remove these that still have an associated rule with.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
@@ -410,7 +410,7 @@ class NSXv3AgentManagerRpcCallBackBase(amb.CommonAgentManagerRpcCallBackBase):
                 timestamp.update()
     
     def first_boot(self):
-        LOG.info("Executing one-time first-boot tasks")
+        LOG.info("One-time first-boot tasks - started ...")
         # A failure in this code should not lead to failure in agent operation
         try:
             os_id_nsx_id_map = self.infra.get_orphan_ipsets()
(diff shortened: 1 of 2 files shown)


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_agent.py
```diff
@@ -393,6 +395,10 @@ def sync_inventory(self):
             timestamp = nsxv3_facada.Timestamp(
                 "last_full_synchronization", self.nsxv3, TransportZones, tz, cfg.CONF.AGENT.sync_full_schedule)
 
+            if self.is_first_boot:
+                self.first_boot()
+                self.is_first_boot = False
+
```
notandy, line 401: Doing the cleanup on startup will cause agent to hold all normal operations till the cleanup is done.
I don't think the cleanup of 50k ipsets is going to be so fast that the agent is done before the agent-liveness-check kicks in that kills the agent.

Please reconsider doing the cleanup in the shallow loop and probably also limit the amount of ipsets to delete.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/nsxv3_policy.py
```diff
@@ -622,6 +622,62 @@ def get_revision(self, resource_container, resource_id):
         else:
             return -1
 
+    def get_orphan_ipsets(self):
```
notandy, line 625: Can we have a test case / are we sure that this code doesn't delete anything in-use?

notandy (changes requested):

PR #62, opened 2021-05-02
Introduce port aware security

tsimchev-vmw:
In the past all openstack security objects have been synced
as part of the agent operation. With this change only security
objects related to ports will be synched.

Additional changes
- Improve orphan object cleanup
- Optimise object creation
- Add unit-tests for management provider
- Add unit-tests end-to-end for Agent Manager
- Mock OpenStack Inventory for RPC
- Mock NSX-T Inventory for REST
- Enable automated cleanup of Policy objects
- Implement automatic switch between Management and Policy APIs
- Implement end-to-end unit tests for Policy API
- Implement CLI support for object synchronization
- Implement proper async operations termination


AUTHORS
```diff
@@ -0,0 +1,11 @@
+Aleksey Nishev <anishev@vmware.com>
+Andrew Karpow <andrew.karpow@sap.com>
+Andrew Karpow <andy@ndyk.de>
+Csaba Seres <45421502+Scsabiii@users.noreply.github.com>
+Johannes Kulik <43781403+joker-official@users.noreply.github.com>
```
joker-at-work, line 5: noreply.github.com mail addresses don't make much sense. Can we remove them?

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
     @log_helpers.log_method_call
-    def get_port_allowed_pairs(self, context, port_id):
-        return db.get_port_allowed_pairs(context, port_id)
+    def get_port(self, context, host, port_id):
+        port = db.get_port(context, host, port_id)
 
-    @log_helpers.log_method_call
-    def get_port_addresses(self, context, port_id):
-        return db.get_port_addresses(context, port_id)
+        # vRA with NSX-T does not support CIDR as port manual binding - skipping X/X
```
joker-at-work, line 221: `vRA`?

networking_nsxv3/common/config.py
```diff
@@ -198,16 +164,11 @@
...
-        help="Remvoe dynamic membership criteria from security groups."
-        # NSX-T objects will be created but not enforced on the segment ports
-    ),
     cfg.IntOpt(
-        'nsxv3_remove_orphan_items_count',
+        'nsxv3_cache_refresh_window',
         default=30,
-        help="Amount of items to be deleted if per sync loop."
-    ),
+        help="Amount of seconds transferred from cache history on refresh"
```
joker-at-work, line 170: Does "transferred" mean "removed" here? If not: transferred where?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -0,0 +1,206 @@
...
+        except Exception:
+            pass
+        LOG.info("Session token - acquired, connected to NSX-T {}".format(self._version))
+
+    def _params(self, **kwargs):
+        kwargs["timeout"] = self._timeout
+        kwargs["url"] = "{}{}".format(self._base_path, kwargs["path"])
+        del kwargs["path"]
+        if self._login_path not in kwargs["url"] and "data" in kwargs:
+            kwargs["data"] = json.dumps(kwargs["data"])
```
joker-at-work, line 165: I'd like to understand why we use the manual `json.dumps()` here and not the `json` argument requests provides. Is it because that would also set the `application/json` content-type?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -0,0 +1,206 @@
...
+
+        try:
+            # Refresh version after login
+            self.version(refresh=True)
+        except Exception:
+            pass
+        LOG.info("Session token - acquired, connected to NSX-T {}".format(self._version))
+
+    def _params(self, **kwargs):
+        kwargs["timeout"] = self._timeout
```
joker-at-work, line 161: What's the reason for setting `timeout` on every request instead of on the session?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -0,0 +1,206 @@
...
+    def version(self, refresh=False):
+        if not self._version or refresh:
+            resp = self.get(path="/api/v1/node/version")
+            if resp.ok:
+                self._version = resp.json()['product_version']
+        return versionutils.convert_version_to_tuple(self._version)
+
+    def _login(self):
+        LOG.info("Session token - acquiring")
+        now = int(datetime.datetime.now().strftime("%s"))
```
joker-at-work, line 134: In case you ever want the same result but with shorter code:
```python
>>> datetime.datetime.now().strftime('%s')
'1620027230'
>>> time.time()
1620027232.8798525
>>> int(time.time())
1620027275
```

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider.py
```diff
@@ -0,0 +1,130 @@
+class Provider(object):
+    """
+    Provider interface used for realization of OpenStack objects
```
joker-at-work, line 3: If this is supposed to be an interface, would it make sense to enforce it via [Abstract Base Class](https://docs.python.org/3/library/abc.html)?

joker-at-work (commented):
Skimmed through the code and have  some questions.

AUTHORS
tsimchev-vmw: Yes, the AUTHORS was generated by the tox. I will remove them manually.

networking_nsxv3/common/config.py
tsimchev-vmw: It actually transferred. When cache rebuilding is triggered there might be active jobs that will modify it, thus the transaction history has to be transferred on top of the regenerated cache.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
tsimchev-vmw: The main reason was to login, as it uses the post but not json. But will make it another way. `json` arguments makes more sense. Thanks

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
tsimchev-vmw: I read about issues with timeout on session objects, did not get deep into actually testing them, thus took the safe way.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py
```diff
@@ -0,0 +1,324 @@
...
+    rpc = manager.get_rpc_callbacks(context=None, agent=None, sg_agent=None)
+
+    for id in sg_ids:
+        rpc.security_groups_member_updated(id)
+    
+    for id in sg_ids:
+        rpc.security_groups_rule_updated(id)
+
+    for id in qos_ids:
+        rpc.update_policy(id)
```
notandy, line 291: It looks like update_policy takes 2 parameters.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py
```diff
@@ -0,0 +1,324 @@
...
+    cfg.CONF.AGENT_CLI for options
+    """
+    LOG.info("VMware NSXv3 Agent CLI")
+    common_config.init(sys.argv[1:])
+    common_config.setup_logging()
+    profiler.setup(nsxv3_constants.NSXV3_BIN, cfg.CONF.host)
+
+    sg_ids = cfg.CONF.AGENT_CLI.neutron_security_group_id
+    port_ids = cfg.CONF.AGENT_CLI.neutron_port_id
+    qos_ids = cfg.CONF.AGENT_CLI.neutron_qos_policy_id
```
notandy, line 279: Can we make the config-options passable via cmd-line?

AUTHORS
```diff
@@ -0,0 +1,11 @@
+Aleksey Nishev <anishev@vmware.com>
+Andrew Karpow <andrew.karpow@sap.com>
+Andrew Karpow <andy@ndyk.de>
```
notandy, line 3: Can you remove my private mail address, thanks.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
tsimchev-vmw: @joker-at-work I did not found a way to set this on a session, could you point me out?

joker-at-work: [Looks like](https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/#setting-default-timeouts) you need to extend the `HTTPAdapter` to change the default behavior. Might be easier not to do this. Sorry, didn't remember that correctly.

networking_nsxv3/api/rpc.py
```diff
@@ -108,90 +81,46 @@ class NSXv3ServerRpcApi(object):
...
+        return cctxt.call(self.context, 'get_ports_revisions',
+                          host=self.host, limit=limit, offset=offset)
 
     @log_helpers.log_method_call
-    def get_qos_policy_revision_tuples(
-            self, limit=_LIMIT, created_after=_CREATE_AFTER):
+    def get_qoses_revisions(self, limit, offset):
         cctxt = self.client.prepare()
-        return cctxt.call(self.context, 'get_qos_policy_revision_tuples',
-                          limit=limit, created_after=created_after)
```
fwiesel, line 137: What is the motivation to go from "created_after/limit" to "offset/limit"?
"Offset" queries are usually less performant and instable over modifications.

networking_nsxv3/api/rpc.py
```diff
@@ -33,32 +31,19 @@ def __init__(self, context):
...
-        )
+            topics.AGENT, nsxv3_constants.NSXV3, topics.UPDATE, host)
 
+        # fanout=True - broadcast to all agents, False only to the host
         return self.rpc.prepare(
             version=self.version,
             topic=topic,
-            fanout=fanout
-        )
+            fanout=True if host is None else False)
```
fwiesel, line 41: Personally, I would prefer a boolean expression directly, e.g.
`fanout=(host is None)`

But that's a matter of taste. Not sure, if we have a convention for that.

networking_nsxv3/api/rpc.py
```diff
@@ -108,90 +81,46 @@ class NSXv3ServerRpcApi(object):
...
         self.client = rpc.get_client(target)
-        self.context = context
-        self.host = host
-        self.topic = topic
+        self.host = cfg.CONF.host        
 
     @log_helpers.log_method_call
-    def get_port_revision_tuples(
-            self, limit=_LIMIT, created_after=_CREATE_AFTER):
+    def get_ports_revisions(self, limit, offset):
```
fwiesel, line 92: I think, the function call should make clear, what the relationship between ports and revisions is.
E.g. `get_ports_with_revisions`, `get_ports_and_revisions` or `get_ports_by_revisions`.

networking_nsxv3/api/rpc.py
```diff
@@ -108,90 +81,46 @@ class NSXv3ServerRpcApi(object):
...
-            host=self.host,
-            limit=limit,
-            created_after=created_after)
+        return cctxt.call(self.context, 'get_ports_revisions',
+                          host=self.host, limit=limit, offset=offset)
 
     @log_helpers.log_method_call
-    def get_qos_policy_revision_tuples(
-            self, limit=_LIMIT, created_after=_CREATE_AFTER):
+    def get_qoses_revisions(self, limit, offset):
```
fwiesel, line 98: I am not a fan of `qoses`, and would suggest to rather to stay with `qos_policies`. Here I also would like some word marking the relationship between QoS-Policies and revisions.

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
         return self._plugin
 
     @log_helpers.log_method_call
-    def get_port_revision_tuples(self, context, host, limit, created_after):
-        return db.get_port_revision_tuples(context, host, limit, created_after)
+    def get_ports_revisions(self, context, host, limit, offset):
```
fwiesel, line 173: Again: Relationship of ports and revisions unclear from name.

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
+        return db.get_ports_revisions(context, host, limit, offset)
 
     @log_helpers.log_method_call
-    def get_qos_policy_revision_tuples(self, context, limit, created_after):
-        return db.get_qos_policy_revision_tuples(context, limit, created_after)
-
+    def get_qoses_revisions(self, context, host, limit, offset):
```
fwiesel, line 177: `qoses` -> `qos_policies`
Relationship of ports and revisions unclear from name.

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
+        return db.get_qoses_revisions(context, host, limit, offset)
+    
     @log_helpers.log_method_call
-    @oslo_messaging.expected_exceptions(exceptions.ObjectNotFound)
-    def get_security_group_revision(self, context, security_group_id):
-        return db.get_security_group_revision(context, security_group_id)
+    def get_security_groups_revisions(self, context, host, limit, offset):
```
fwiesel, line 181: Relationship of Security-Groups and revisions unclear from name.

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
-        return db.has_security_group_tag(
-            context, security_group_id, tag_name)
+    def get_security_group(self, context, security_group_id):
+        id_rev = db.get_security_group_revision(context, security_group_id)
+        if id_rev:
+            return {
+                "id": id_rev[0],
+                "revision_number": id_rev[1],
+                "tags": db.get_security_group_tag(context, security_group_id),
+                "ports": db.get_port_id_by_sec_group_id(context, sec_group_id)
```
fwiesel, line 192: I can't find `sec_group_id` in the body of the function, shouldn't it be `security_group_id``?

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
+    def get_rules_for_security_groups_id(self, context, security_group_id):
+        return db.get_rules_for_security_groups_id(context, security_group_id)
 
     @log_helpers.log_method_call
-    def get_qos_bwl_rules(self, context, qos_id):
-        return db.get_qos_bwl_rules(context, qos_id)
+    def get_security_group_members_ips(self, context, security_group_id):
+        return \
+            db.get_security_group_members_ips(context, security_group_id) + \
+            db.get_security_group_members_address_bindings_ips(context, security_group_id)
```
fwiesel, line 203: I am a bit conflicted about that call, as  it is named exactly like the first db api call, but also does a second api call.
I would suggest renaming it.

I am suprised though that you have to do two independent db api calls to get all the IPs you want. I have to look deeper into the calls to understand it to make a more qualified comment. Personally, I think there should be a single one, but that may be a topic for a later change.

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
-    @log_helpers.log_method_call
-    def get_port_addresses(self, context, port_id):
-        return db.get_port_addresses(context, port_id)
+        # NSX-T does not support CIDR as port manual binding - skipping X/X
 
-    @log_helpers.log_method_call
-    def get_rules_for_security_groups_id(self, context, security_group_id):
-        return db.get_rules_for_security_groups_id(context, security_group_id)
+        for ip in db.get_port_addresses(context, port_id):
+            if "/" in ip:
```
fwiesel, line 224: I am not sure, what you are trying to do here. Are you trying to filter out netmasks? What about "/32" (in ipv4)?
Can't you filter by checking for  the mac address to achieve the same?

Probably a topic for a later change, but I think any filtering should happen via a database query.

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +170,82 @@ def plugin(self):
...
+                continue
+            port["address_bindings"].append({"ip_address": ip,  "mac_address": port["mac_address"]})
 
-    @log_helpers.log_method_call
-    def get_security_group_members_ips(self, context, security_group_id):
-        return db.get_security_group_members_ips(context, security_group_id)
+        for ip, mac in db.get_port_allowed_pairs(context, port_id):
+            if "/" in ip:
```
fwiesel, line 229: Same as with https://github.com/sapcc/networking-nsx-t/pull/62/files#r626504018

networking_nsxv3/common/config.py
```diff
@@ -149,7 +114,8 @@
     ),
     cfg.StrOpt(
         'nsxv3_login_password',
-        default='VMware1!',
+        default='VMwarensbu_1!',
```
fwiesel, line 117: I would suggest to leave the default empty.

networking_nsxv3/common/synchronization.py
```diff
@@ -187,9 +184,19 @@ def start(self):
 
     def stop(self):
         """ Gracefully terminates the runner instance """
+        while True:
```
fwiesel, line 187: That doesn't look right to me: You are polling the `active` and `passive` queue, but do not seem to set any flag/condition to avoid new work being put into the queues.
Is that done externally? I would like that the termination logic is self-contained in the `stop()` routine.

networking_nsxv3/common/synchronization.py
```diff
@@ -142,33 +142,30 @@ def run(self, priority, ids, fn):
...
                 if priority.value == Priority.HIGHEST:
                     self._active.put_nowait(job)
                 else:
                     self._passive.put_nowait(job)
             except eventlet.queue.Full as err:
-                LOG.error(MESSAGE.format(fn.__name__, jid, priority, err))
+                LOG.error(MESSAGE.format(err, jid, priority.name, fn.__name__))
 
     def _start(self):
         while True:
```
fwiesel, line 156: I would expect that there is some way to actually stop the loop.

fwiesel (changes requested):
I think, the Runner code requires some rework regarding the termination.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
```diff
@@ -0,0 +1,824 @@
...
+        return meta
+
+    def metadata_refresh(self, resource_type):
+        if resource_type == Provider.SG_RULE:
+            return # Not cached
+        
+        resources = self._cache.get(resource_type)
+        backup = dict()
+        resources["backup"] = backup
+        eventlet.sleep(self._chache_refresh_window)
```
notandy, line 514: Wouldn't this cause a delay of 30 seconds for every call to metadata_refresh (which is called by outdated())? Not sure this is intended instead - for example to just do the refresh only every 30 seconds - else return.

networking_nsxv3/api/rpc.py
tsimchev: It was mainly due to the performance (thought that offset has better performance) and unification (every query can have limit and offset, but not true for create_after, or at least for some it does not make sense).

I see your point and it makes sense. Will roll that back.

- [ ] Replace offset with create after

networking_nsxv3/api/rpc.py
tsimchev: - [ ] Research how to use single query for SG members all IPs

networking_nsxv3/api/rpc.py
tsimchev: NSX-T does not support CIDR as port manual binding - skipping X/X, but for /32 it makes sense.

- [ ] Remove /32 from CIDR and use the base IP

networking_nsxv3/common/synchronization.py
tsimchev: The shutdown logic is applicable only to the case when the agent is run as CLI with stopped synchronization. Then I am expecting termination after all of the work have been completed. I left timeout termination to external system, as I cannot determine when is the best time to kill an active synchronization process.

networking_nsxv3/api/rpc.py
notandy: This is strange change is taken over by a commit from me, here the explanation:

Initial implementation of policy api used the the addresses collected from all remote security group hosts directly, Whereas mgmt api used netaddr python module to collapse continuous addresses to address codes. We hit without the "collapse" the limit of ips per group in one region with one customer, so I adapted the code to use the same collapse code that mgmt api is using.

Funny thing now: NSXT handles cidr independently from pure up addresses (1.2.3.4 and 1.2.3.4/32 are two distinct ip addresses in the group, LOLWTF!) , so I added the strange check for /32 that leaves the existing host addresses instead introducing duplicates.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
tsimchev: Outdated => metadata_refresh for a resource type is called once on every synchronization. As the metadata is rebuilt by querying NSX-T multiple times to get the full state, there is a change to miss some ongoing create requests that might lead to ambiguosity. This timeout acts as a transaction log interval in which all active jobs must be completed.
The value of 30 seconds can be adjusted based on the size of NSX-T managers and the expected load.

networking_nsxv3/api/rpc.py
tsimchev: @notandy I read this as I can remove "/" from here and relay on the code bellow, correct?

def cidr(self, os_cidrs):
        return [str(ip).replace("/32", "") for ip in netaddr.IPSet(os_cidrs).iter_cidrs()]

notandy: True, should be fine

networking_nsxv3/api/rpc.py
tsimchev: @fwiesel merging both queries into a single database query will be simply concatenation of two results. I do not see any benefit to make this change. Let me know if you see this differently.

fwiesel: Combining queries into one is indeed more a minor improvement. It saves a single round-trip, so fractions of milliseconds. And the concatenation will be done in the DB instead in python, which should be a bit faster. I would not say, that needs to be done, just a suggestion.

My main trouble is the naming. It is named the same as the first db api call, but it does something different, as does a second api call to the db. I would prefer a different name reflecting the differences.

networking_nsxv3/api/rpc.py
fwiesel: I am not so happy about that. the function is named `cidr`, so returning a cidr does make sense.
I would suggest to add simply a function to filter out hosts, and then check for either no netmask, or host-netmask.
Ideally, we also should check for IPv6, or does NSX-T does not IPv6?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
fwiesel: I think, the point is more that there is a `sleep` inside the function, but if you want do it timed, it would be more sensible to do that on the caller side of the function.
This way I might call this function in an RPC call, and that RPC call will timeout on the RPC-timeout, because it is potentially lower than  the value `_chache_refresh_window`.

networking_nsxv3/common/synchronization.py
fwiesel: I think, I didn't manage to get my point across here. It is essentially the same as: https://github.com/sapcc/networking-nsx-t/pull/62#discussion_r626828874

networking_nsxv3/common/synchronization.py
fwiesel: That would be fine, but personally, I would prefer an explicit variable, e.g. something like `_terminating`. But is that sufficient? I mean, isn't there work done in the loop which needs to be finished before terminating? It's an honest question, I haven't understood the code yet to that extend.

networking_nsxv3/api/rpc.py
tsimchev: 200: will be renamed to `get_security_group_members_effective_ips`

networking_nsxv3/api/rpc.py
tsimchev: @fwiesel, I will correct the function to use the proper name for its purpose. Note that `netaddr.IPSet(os_cidrs).iter_cidrs()` works also for IPv6 addresses. And correct with `/32` we are doing small optimisation for IPv4 only, will make this optimisation only if the address is IPv4.
def get_compacted_cidrs(self, os_cidrs):
        """
        Reduce number of CIDRs based on the netmask overlapping
        """

networking_nsxv3/common/synchronization.py
tsimchev: The pool size can get to zero only if a termination operation is triggered. According to the documentation all jobs needs to finish prior the size is set to 0, thus it will be safe to use this as a termination trigger.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
tsimchev: I see your point and it makes sense. I do not expect this call to be ever made in RPC context.
I would prefer to leave it that way until there is a need.

networking_nsxv3/common/synchronization.py
tsimchev: - [ ] Prevent Runner to take new job when not in state "started"

.github/workflows/unittests.yaml
```diff
@@ -0,0 +1,36 @@
+  # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Unit Tests
+
+on:
+  push:
+    branches: [ stable/quuens-m3 ]
```
fwiesel, line 8: Spelling mistake, I assume

networking_nsxv3/api/rpc.py
fwiesel: That is still open, correct?

networking_nsxv3/api/rpc.py
```diff
@@ -108,110 +81,77 @@ class NSXv3ServerRpcApi(object):
...
-    @log_helpers.log_method_call
-    def get_port_security_groups(self, port_id):
-        cctxt = self.client.prepare()
-        return cctxt.call(self.context, 'get_port_security_groups',
-                          port_id=port_id)
+        return cctxt.call(self.context, 'get_port', host=self.host, port_id=port_id)
 
     @log_helpers.log_method_call
-    def get_port_allowed_pairs(self, port_id):
+    def get_rules_for_security_groups_id(self, security_group_id):
```
fwiesel, line 126: `get_rules_for_security_groups_id` should be `get_rules_for_security_group_id` (singular group id)

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def get_compacted_cidrs(self, os_cidrs):
+        """
+        Reduce number of CIDRs based on the netmask overlapping
+        """
+        compacted_cidrs = []
+        for cidr in netaddr.IPSet(os_cidrs).iter_cidrs():
+            if cidr.version == 4 and cidr.prefixlen == 32:
+                compacted_cidrs.append(str(cidr.ip))
+            else:
```
fwiesel, line 35: What about `cidr.version == 6 and cidr.prefixlen == 128`?

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 2.1
-envlist = py{35,27},functional,pep8,pip-missing-reqs
```
fwiesel, line 3: Don't we need to test on python 3 anymore? I assume we want to go on a later Openstack version soon, which only works on python3

fwiesel (changes requested):
Main issue I have is the "limit,offset" calls. Otherwise it looks good to me

tox.ini
tsimchev: Correct, there is another branch and pull request for Ussuri with python 3.6.9 (changes from both branches are in sync)
Here we use only python 2.7 due to some breaking changes.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
tsimchev: - [x] Add IPv6 host CIDER normalization

networking_nsxv3/api/rpc.py
tsimchev: - [x] Improve method naming to better reflect its return result

.github/workflows/unittests.yaml
tsimchev: - [x] Fix gitactions typo

networking_nsxv3/common/config.py
tsimchev: - [x] Remove default password

networking_nsxv3/api/rpc.py
tsimchev: - [x] Improve CIDR normalization logic

networking_nsxv3/api/rpc.py
tsimchev: - [x] Improve DB query precision replacing offset with created_after

fwiesel (changes requested):
Since you changed the queries back to `created_after`, we need the queries to be ordered again by that timestamp.

But I was not looking into the database schema, and it looks like we have a bit of trouble there.
The `created_at` attribute is in a different table called `standardattributes` and on top of that, there is no index on `created_at`... That means the query will very inefficient.

Going over the primary key is also not advisable, as it is a uuid,  which is practically random.
However, we could go over `standard_attr_id`, which is auto-increment field.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -161,9 +153,12 @@ def _realize(self, resource_type, delete, convertor, os_o, provider_o):
         meta = self.metadata(resource_type, os_id)
         if meta:
             if delete:
+                LOG.info(report, "deleted")
```
fwiesel, line 156: Should that really stay in? It looks like debug code to me. Honest question, I do not know

fwiesel (approved):
Looks good to me

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -0,0 +1,217 @@
...
+    def _login(self):
+        LOG.info("Session token - acquiring")
+        now = int(time.time())
+        with LockManager.get_lock(self._base_path):
+            if now > self._login_timestamp:
+                resp = requests.post(**self._params(path=self._login_path,
+                                                    data=self._login_data,
+                                                    verify=self._session.verify))
+                if resp.status_code != requests.codes.ok:
+                    resp.raise_for_status()
```
notandy, line 143: requests.codes.ok is a try/catch wrapper around raise_for_status(), so this can be replaced with `resp.raise_for_status()`

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
```diff
@@ -0,0 +1,273 @@
...
+        # Initializing metadata
+        self.all(dryrun=True)
+
+    def _os_meta(self, query):
+        step = cfg.CONF.AGENT.rpc_max_records_per_query
+        cursor = 0
+        meta = dict()
+        while cursor != -1:
+            result = query(step, cursor)
+            meta.update({k:v for (k,v) in result})
```
notandy, line 31: this implies that result entities consists of a tuple of 2 elements, but for example `get_ports_with_revisions` return a list of 3-tuples, causing this error: 
```python
Traceback (most recent call last):
  File "/Users/d072895/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/211.7142.13/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevd.py", line 1491, in _exec
    func()
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py", line 354, in main
    NSXv3Manager(rpc=nsxv3_rpc.NSXv3ServerRpcApi()),
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py", line 116, in __init__
    legacy_provider=legacy_provider)
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py", line 23, in __init__
    self.all(dryrun=True)
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py", line 65, in all
    port_meta = self._os_meta(r.get_ports_with_revisions)
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py", line 31, in _os_meta
    meta.update({k:v for (k,v) in result})
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py", line 31, in <dictcomp>
    meta.update({k:v for (k,v) in result})
ValueError: too many values to unpack
```

notandy (changes requested):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
notandy: Screenshot 2021-05-21 at 14 05 13

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py
```diff
@@ -0,0 +1,362 @@
...
+    def get_agent_configurations(self):
+        """Establishes the agent configuration map.
+        The content of this map is part of the agent state reports to the
+        neutron server.
+        :return: map -- the map containing the configuration values
+        :rtype: dict
+        """
+        c = cfg.CONF.NSXV3
+        return {
+            'nsxv3_policy_migration_limit': c.nsxv3_policy_migration_limit,
```
notandy, line 196: causing
```
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service [req-bd0b572e-64e4-48d0-a859-febc9021542a - - - - -] Error starting thread.: NoSuchOptError: no such option nsxv3_policy_migration_limit in group [NSXV3]
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service Traceback (most recent call last):
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/oslo_service/service.py", line 729, in run_service
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     service.start()
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/osprofiler/profiler.py", line 159, in wrapper
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     result = f(*args, **kwargs)
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/neutron/plugins/ml2/drivers/agent/_common_agent.py", line 89, in start
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     configurations.update(self.mgr.get_agent_configurations())
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py", line 196, in get_agent_configurations
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     'nsxv3_policy_migration_limit': c.nsxv3_policy_migration_limit,
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/oslo_config/cfg.py", line 3378, in __getattr__
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     return self._conf._get(name, self._group)
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/oslo_config/cfg.py", line 2940, in _get
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     value = self._do_get(name, group, namespace)
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/oslo_config/cfg.py", line 2957, in _do_get
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     info = self._get_opt_info(name, group)
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service   File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/oslo_config/cfg.py", line 3114, in _get_opt_info
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service     raise NoSuchOptError(opt_name, group)
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service NoSuchOptError: no such option nsxv3_policy_migration_limit in group [NSXV3]
2021-05-21 15:50:53,812.812 64393 ERROR oslo_service.service 
```

notandy (changes requested):

notandy:
```python
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server [req-dbf05411-e8b9-4e17-8d05-bbad299d0f8a - - - - -] Exception during message handling: MultipleResultsFound: Multiple rows were found for one_or_none()
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server Traceback (most recent call last):
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/local/lib/python2.7/site-packages/oslo_messaging/rpc/server.py", line 163, in _process_incoming
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     res = self.dispatcher.dispatch(message)
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/local/lib/python2.7/site-packages/oslo_messaging/rpc/dispatcher.py", line 220, in dispatch
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     return self._do_dispatch(endpoint, method, ctxt, args)
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/local/lib/python2.7/site-packages/oslo_messaging/rpc/dispatcher.py", line 190, in _do_dispatch
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     result = func(ctxt, **new_args)
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/local/lib/python2.7/site-packages/oslo_log/helpers.py", line 67, in wrapper
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     return method(*args, **kwargs)
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/src/networking-nsxv3/networking_nsxv3/api/rpc.py", line 221, in get_port
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     port = db.get_port(context, host, port_id)
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/src/networking-nsxv3/networking_nsxv3/db/db.py", line 176, in get_port
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     trunk_model.Trunk
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2816, in one_or_none
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server     "Multiple rows were found for one_or_none()")
neutron-server-857ccc4b-5pr8k neutron-server 2021-05-21 13:56:29,599.599 23 ERROR oslo_messaging.rpc.server MultipleResultsFound: Multiple rows were found for one_or_none()
```

networking_nsxv3/db/db.py
```diff
@@ -167,11 +158,14 @@ def get_port(context, port_id):
         PortBinding.host,
         PortBinding.vif_details,
         trunk_model.Trunk.port_id
+    ).select_from(
+        Port
     ).filter(
-        Port.id == port_id
+        Port.id == port_id,
+        PortBindingLevel.host == host
```
notandy, line 165: You probably meanth PortBinding.host, because joining with PortBindingLevel joins with all levels of this host. (Causing to return an excessive amount of ports)

notandy (changes requested):

notandy (changes requested):

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +171,83 @@ def plugin(self):
...
 
-    @log_helpers.log_method_call
-    def get_port_addresses(self, context, port_id):
-        return db.get_port_addresses(context, port_id)
+        # NSX-T does not support CIDR as port manual binding - skipping X/X
 
-    @log_helpers.log_method_call
-    def get_rules_for_security_groups_id(self, context, security_group_id):
-        return db.get_rules_for_security_groups_id(context, security_group_id)
+        for ip in db.get_port_addresses(context, port_id):
```
notandy, line 225: get_port_addresses() returns a list of [ip_address, subnet_id], which causes later in the Ports call an invalid structure.

If the subnet_id is not needed, maybe completely remove from rpc?

notandy (changes requested):

networking_nsxv3/api/rpc.py
notandy: e.g. 
```
2021-05-21 17:26:30,872.872 86289 ERROR networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.client_nsx [-] Request=Function PUT Argumetns {'path': '/api/v1/logical-ports/a4a2d2f3-c9cd-4780-92f1-4f1f5ed54347', 'data': {'switching_profile_ids': [{'key': 'SpoofGuardSwitchingProfile', 'value': u'f833742f-459f-4783-b20c-9bd61752675c'}, {'key': 'IpDiscoverySwitchingProfile', 'value': u'b73982a9-c0ea-4632-a62f-d4b831ac4598'}], 'tags': [{'scope': 'security_group', 'tag': u'4360da8e-b208-4bcf-8021-7a773d795f84'}, {'scope': 'age', 'tag': 1621610790}, {'scope': 'agent_id', 'tag': 'nsxm-l-01a.corp.local'}, {'scope': 'revision_number', 'tag': 10}], 'address_bindings': [{u'ip_address': [u'10.180.0.17', u'858f0d2d-5622-4be2-abe6-e5a286bde3b7'], u'mac_address': u'fa:16:3e:49:44:ae'}], 'logical_switch_id': u'91d1e26c-7c03-493b-951f-b7f88f6fa6eb', 'admin_state': 'UP', 'attachment': {'attachment_type': 'VIF', 'context': {'vif_type': 'PARENT', 'traffic_tag': 2054, 'resource_type': 'VifAttachmentContext'}, 'id': u'e46fc1b8-27bd-4353-bd44-b18f57df6baa'}, '_revision': 194, 'display_name': u'e46fc1b8-27bd-4353-bd44-b18f57df6baa'}} Response=Error Code=400 Message={
  "httpStatus" : "BAD_REQUEST",
  "error_code" : 208,
  "module_name" : "common-services",
  "error_message" : "The property address_bindings.null.ip_address must be a valid string"
}
```

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
tsimchev: I will change the priority.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
```diff
@@ -0,0 +1,880 @@
...
+                    'IPv6': "ICMPv6"
+                }.get(ethertype)
+            }, None)
+
+        if protocol in ["tcp", "udp"]:
+            return ({
+                "resource_type": "L4PortSet{}".format(subtype),
+                "l4_protocol": {'tcp': "TCP", 'udp': "UDP"}.get(protocol),
+                "destination_ports": ["{}-{}".format(min, max) \
+                    if min != max and max else str(min)],
```
notandy, line 447: This causes an Error if an user creates rules with without min/max destination port or port = 0:
https://docs.openstack.org/api-ref/network/v2/?expanded=create-security-group-rule-detail#create-security-group-rule
Which causes min and max to be Null.

```python
  "details" : "Field level validation errors: {value 'None' of property rules[287].service_entries[0].destination_ports[0] violates format 'port-or-range'}",
  "httpStatus" : "BAD_REQUEST",
  "error_code" : 255,
  "module_name" : "common-services",
  "error_message" : "Field level validation errors: {value 'None' of property rules[287].service_entries[0].destination_ports[0] violates format 'port-or-range'}"
```

Possible fix:
```python
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
@@ -440,11 +440,14 @@ class Payload(object):
             }, None)

         if protocol in ["tcp", "udp"]:
+            destination_ports = "1-65535"
+            if min or max:
+                destination_ports = "{}-{}".format(min, max) \
+                    if min != max and max else str(min)
             return ({
                 "resource_type": "L4PortSet{}".format(subtype),
                 "l4_protocol": {'tcp': "TCP", 'udp': "UDP"}.get(protocol),
-                "destination_ports": ["{}-{}".format(min, max) \
-                    if min != max and max else str(min)],
+                "destination_ports": [destination_ports],
                 "source_ports": ["1-65535"]
             }, None)
```

notandy (changes requested):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
```diff
@@ -0,0 +1,880 @@
...
+                "id": os_port.get("id"),
+                "context": {
+                    "resource_type": "VifAttachmentContext",
+                    "vif_type": "PARENT",
+                    "traffic_tag": \
+                        p.get("vif_details").get("segmentation_id")
+                }
+            },
+            "tags": self.tags(os_port, more={NSXV3_SECURITY_GROUP_SCOPE:os_id \
+                for os_id in p.get("security_groups")})
```
notandy, line 281: Assigning multiple security-groups per port doesn't work. With a dictionary, the static KEY (NSXV3_SECURITY_GROUP_SCOPE=security_group) can only exist once. Therefor, it will be always replaced with the last security group id in the array.

notandy (changes requested):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
tsimchev: Fixed.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def sg_rules_container(self, os_sg, provider_sg):
+        return {
+            "category": "Application",
+            "display_name": os_sg.get("id"),
+            "stateful": True,
+            "tcp_strict": NSXV3_CAPABILITY_TCP_STRICT in os_sg.get("tags"),
+            "scope": [
+                "/infra/domains/default/groups/{}".format(
+                    provider_sg.get("scope"))
```
notandy, line 76: I am a bit worried about the intense use of .get() on dictionaries:

For some critical values, like the scope, a missing key could hide an critical error and eventually cause issues by applying a security group to all ports. This is only hypothetical (I didn't check if thats the case in this example), but for absolutely required values as here I don't see why not just use `[key]`, which would make issues visible and abort the operation by throwing an exception.

notandy (changes requested):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
tsimchev: I am using it in order to create the full payload, then on error to know how it looked like.
In this case the request will fail due to mall formatted path.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py
tsimchev: Fixed

networking_nsxv3/api/rpc.py
tsimchev: Fixed

networking_nsxv3/api/rpc.py
```diff
@@ -231,69 +171,83 @@ def plugin(self):
...
 
-    @log_helpers.log_method_call
-    def get_security_group_members_ips(self, context, security_group_id):
-        return db.get_security_group_members_ips(context, security_group_id)
+        for ip, mac in db.get_port_allowed_pairs(context, port_id):
+            if "/" in ip:
+                continue
+            port["address_bindings"].append({"ip_address": ip, "mac_address": mac})
+
+        for sg_id,_ in db.get_port_security_groups(context, port_id):
```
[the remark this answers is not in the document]
tsimchev: Fixed

networking_nsxv3/db/db.py
tsimchev: Fixed

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py
tsimchev: Fixed

notandy:
Currently, 3.0 E2E Tests fails with:

```python
traceback-1: {{{
Traceback (most recent call last):
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/tests/functional/test_realization_brownfield.py", line 52, in tearDown
    self.cleanup()
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/tests/functional/test_realization_brownfield.py", line 48, in cleanup
    self.assertEquals(meta["meta"], dict())
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/.tox/functional/lib/python2.7/site-packages/testtools/testcase.py", line 411, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/.tox/functional/lib/python2.7/site-packages/testtools/testcase.py", line 498, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: !=:
reference = {u'00000000-0000-0000-0000-000000000001': {'_revision': 0,
                                           'age': u'1622403880',
                                           'id': u'00000000-0000-0000-0000-000000000001',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000002': {'_revision': 0,
                                           'age': u'1622403881',
                                           'id': u'00000000-0000-0000-0000-000000000002',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000003': {'_revision': 0,
                                           'age': u'1622403881',
                                           'id': u'00000000-0000-0000-0000-000000000003',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000004': {'_revision': 0,
                                           'age': u'1622403882',
                                           'id': u'00000000-0000-0000-0000-000000000004',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000005': {'_revision': 0,
                                           'age': u'1622403883',
                                           'id': u'00000000-0000-0000-0000-000000000005',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000006': {'_revision': 0,
                                           'age': u'1622403883',
                                           'id': u'00000000-0000-0000-0000-000000000006',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000007': {'_revision': 0,
                                           'age': u'1622403884',
                                           'id': u'00000000-0000-0000-0000-000000000007',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000008': {'_revision': 0,
                                           'age': u'1622403885',
                                           'id': u'00000000-0000-0000-0000-000000000008',
                                           'rev': None},
 u'00000000-0000-0000-0000-000000000009': {'_revision': 0,
                                           'age': u'1622403886',
                                           'id': u'00000000-0000-0000-0000-000000000009',
                                           'rev': None},
 u'22D4CB40-31A6-4C61-A527-76B7867E221B': {'_revision': 0,
                                           'age': u'1622403865',
                                           'id': u'22D4CB40-31A6-4C61-A527-76B7867E221B',
                                           'rev': u'0'},
 u'28778C62-C22F-47DD-801F-CF06DF3D07AD': {'_revision': 0,
                                           'age': u'1622403862',
                                           'id': u'28778C62-C22F-47DD-801F-CF06DF3D07AD',
                                           'rev': u'0'},
 u'34B87931-F273-4C6D-96D0-B3979E30254A': {'_revision': 1,
                                           'age': u'1622403889',
                                           'id': u'34B87931-F273-4C6D-96D0-B3979E30254A',
                                           'rev': u'0'},
 u'ED75FC68-69BB-4034-A6E9-A7586792B229': {'_revision': 2,
                                           'age': u'1622403901',
                                           'id': u'ED75FC68-69BB-4034-A6E9-A7586792B229',
                                           'rev': u'0'},
 u'EDE7338F-9AE3-445C-96D3-D8EDDEBC8277': {'_revision': 0,
                                           'age': u'1622403862',
                                           'id': u'EDE7338F-9AE3-445C-96D3-D8EDDEBC8277',
                                           'rev': u'0'}}
actual    = {}
}}}

Traceback (most recent call last):
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/.tox/functional/lib/python2.7/site-packages/neutron/tests/base.py", line 181, in func
    return f(self, *args, **kwargs)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/tests/functional/test_realization_brownfield.py", line 92, in test_end_to_end
    self._assert_update(c, env)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/tests/functional/test_realization_brownfield.py", line 158, in _assert_update
    self.assertEquals(c.SECURITY_GROUP_AUTH["id"] in m[p.SG_MEMBERS]["meta"], False)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/.tox/functional/lib/python2.7/site-packages/testtools/testcase.py", line 411, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/.tox/functional/lib/python2.7/site-packages/testtools/testcase.py", line 498, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: True != False
```

and 2.4.2 E2E fails with:
```python
2021-05-30 19:45:17.075 1035 INFO networking_nsxv3.tests.environment [-] Environment:Cleanup final state of NSX-T inventory: {
    "Security Group (Rules Enforcement)": {
        "endpoint": "/api/v1/ns-groups", 
        "meta": {}, 
        "meta_transaction": {}
    }, 
    "Security Group (Rules Remote IP Prefix)": {
        "endpoint": "/api/v1/ip-sets", 
        "meta": {
            "::/0": {
                "id": "af47e897-b816-45bd-bbef-a466acc21dcc", 
                "age": "1621937890", 
                "_revision": 0, 
                "rev": null
            }, 
            "::/96": {
                "id": "728affe0-b4f5-41b9-ac7e-568fb634ccb3", 
                "age": "1621937891", 
                "_revision": 0, 
                "rev": null
            }, 
            "0.0.0.0/0": {
                "id": "a983c6d7-6512-4927-a5fe-af16b060cc06", 
                "age": "1621937890", 
                "_revision": 0, 
                "rev": null
            }, 
            "0.0.0.0/24": {
                "id": "4684cd8c-5718-4874-82a0-f12399309850", 
                "age": "1621937890", 
                "_revision": 0, 
                "rev": null
            }
        }, 
        "meta_transaction": {}
    }, 
    "Port": {
        "endpoint": "/api/v1/logical-ports", 
        "meta": {}, 
        "meta_transaction": {}
    }, 
    "Security Group (Members)": {
        "endpoint": "/api/v1/ip-sets", 
        "meta": {}, 
        "meta_transaction": {}
    }, 
    "Security Group (Rules)": {
        "endpoint": "/api/v1/firewall/sections", 
        "meta": {}, 
        "meta_transaction": {}
    }, 
    "Network": {
        "endpoint": "/api/v1/logical-switches", 
        "meta": {
            "1000": {
                "id": "c061859d-5911-41d4-9e90-bb48c9f6f85b", 
                "age": null, 
                "_revision": 0, 
                "rev": null
            }, 
            "3200": {
                "id": "4d6cf8af-a345-4982-b625-437bea2f3603", 
                "age": null, 
                "_revision": 0, 
                "rev": null
            }
        }, 
        "meta_transaction": {}
    }, 
    "QoS": {
        "endpoint": "/api/v1/switching-profiles", 
        "meta": {}, 
        "meta_transaction": {}
    }
}
}}}

Traceback (most recent call last):
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/.tox/functional/lib/python2.7/site-packages/neutron/tests/base.py", line 181, in func
    return f(self, *args, **kwargs)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/tests/functional/test_realization_brownfield.py", line 83, in test_end_to_end
    self._pollute(env, index)
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/tests/functional/test_realization_brownfield.py", line 211, in _pollute
    p.client.put(path=api.POLICY.format(id), data=pp.sg_rules_container({"id": id}, {"rules": [], "scope": id}))
  File "/tmp/build/9b0aeb78/queens-networking-nsx-t.git/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py", line 89, in decorator
    raise Exception(msg, last_err)
Exception: (None, 'Error Code=400 Message={\n  "httpStatus" : "BAD_REQUEST",\n  "error_code" : 287,\n  "module_name" : "common-services",\n  "error_message" : "Json de-serialization error: property scope is unrecognized."\n}')
```

notandy:
Hi,

It seems I found a bug in NSX-T 3.0.1 with following Services API call:

```python
len(self.client.get_all(path=API.SERVICES, params={"default_service": True}))
  >>> 409
len(self.client.get_all(path=API.SERVICES, params={"default_service": False}))
  >>> 409
```

I will add a patch that additionally filters the default service objects.

notandy:
I am currently debugging a bit why the functional test with NSX-T 3.0.1 fails and found following:

When reaching Line 83 in `test_realization_brownfield.py`, a _sync_all refresh is called:

```python
2021-06-02 12:52:20.682 52849 INFO networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.realization [-] Sanitizing provider based on age cycles
```

Which adds the ip-set `22D4CB40-31A6-4C61-A527-76B7867E221B` for deletion:

```python
2021-06-02 12:52:23.427 52849 INFO networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.provider_nsx_mgmt [-] Fetching NSX-T metadata for Type:Security Group (Members).
2021-06-02 12:52:23.503 52849 INFO networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.provider_nsx_mgmt [-] The number of outdated resources for Type:Security Group (Members) Is:5.
2021-06-02 12:52:23.505 52849 INFO networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.realization [-] Realizing 5/5 resources of Type:Security Group Members (Legacy)
2021-06-02 12:52:23.508 52849 INFO networking_nsxv3.common.synchronization [-] Enqueued Resource ID:28778C62-C22F-47DD-801F-CF06DF3D07AD with Priority:HIGH for action legacy_sg_members_realize
2021-06-02 12:52:23.511 52849 INFO networking_nsxv3.common.synchronization [-] Enqueued Resource ID:ED75FC68-69BB-4034-A6E9-A7586792B229 with Priority:HIGH for action legacy_sg_members_realize
2021-06-02 12:52:23.513 52849 INFO networking_nsxv3.common.synchronization [-] Enqueued Resource ID:22D4CB40-31A6-4C61-A527-76B7867E221B with Priority:HIGH for action legacy_sg_members_realize
2021-06-02 12:52:23.515 52849 INFO networking_nsxv3.common.synchronization [-] Enqueued Resource ID:EDE7338F-9AE3-445C-96D3-D8EDDEBC8277 with Priority:HIGH for action legacy_sg_members_realize
2021-06-02 12:52:23.518 52849 INFO networking_nsxv3.common.synchronization [-] Enqueued Resource ID:34B87931-F273-4C6D-96D0-B3979E30254A with Priority:HIGH for action legacy_sg_members_realize
```

And then tries to delete it shortly after it:

```python
2021-06-02 12:52:25.019 52849 INFO networking_nsxv3.common.synchronization [-] Processing Resource ID:22D4CB40-31A6-4C61-A527-76B7867E221B with Priority:HIGH for action legacy_sg_members_realize
```

which fails:

```python
2021-06-02 13:23:39.998 52849 ERROR networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.client_nsx [-] Request=Function DELETE Argumetns {'path': '/api/v1/ip-sets/f645b9ad-204d-4772-a84c-3f0fe9631487', 'params': {}} Response=Error Code=400 Message={
  "httpStatus" : "BAD_REQUEST",
  "error_code" : 3022,
  "module_name" : "internal-framework",
  "error_message" : "One or more relationships exist for object with id IPSet/f645b9ad-204d-4772-a84c-3f0fe9631487."
}
Traceback (most recent call last):
  File "/Users/d072895/.virtualenvs/queens_python2.7/lib/python2.7/site-packages/eventlet/greenthread.py", line 214, in main
    result = function(*args, **kwargs)
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py", line 148, in legacy_sg_members_realize
    l.sg_members_realize({"id": id}, delete=True)
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py", line 762, in sg_members_realize
    self.payload.sg_members_container, sg, dict())
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_mgmt.py", line 645, in _realize
    self.client.delete(path=path, params=params)
  File "/Users/d072895/Workspace/queens/networking-nsx-t/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py", line 89, in decorator
    raise Exception(msg, last_err)
Exception: (None, 'Error Code=400 Message={\n  "httpStatus" : "BAD_REQUEST",\n  "error_code" : 3022,\n  "module_name" : "internal-framework",\n  "error_message" : "One or more relationships exist for object with id IPSet/f645b9ad-204d-4772-a84c-3f0fe9631487."\n}')
```

This happens because the IPSet `22D4CB40-31A6-4C61-A527-76B7867E221B` is used as a remote ip-set by the Section  `ED75FC68-69BB-4034-A6E9-A7586792B229` and it's Rule `70AECE1F-B4AE-4142-9D3E-66CACB714398` which is still realizied.

The cleanup function

```python
    def sg_members_realize(self, sg, delete=False):
        if delete and self.metadata(Provider.SG_RULES, sg.get("id")):
            LOG.debug("Resource:%s with ID:%s deletion is rescheduled due to dependency.",
                      Provider.SG_MEMBERS, sg.get("id"))
            return
        return self._realize(Provider.SG_MEMBERS, delete,
                             self.payload.sg_members_container, sg, dict())
```
Detects if a Rule with the name of the IP-Set is existing in the metadata-cache, and if not deletes it. In this case, the remote-ipset has a different name than the rule that is referencing it. This is a result of the coverage.py inventory. I think, in our environment the remote-ipset has always the same `display_name` as the rule, so this check should be sufficient for skipping the illegal-delete?

tsimchev:
> Detects if a Rule with the name of the IP-Set is existing in the metadata-cache, and if not deletes it. In this case, the remote-ipset has a different name than the rule that is referencing it. This is a result of the coverage.py inventory. I think, in our environment the remote-ipset has always the same `display_name` as the rule, so this check should be sufficient for skipping the illegal-delete?

@notandy In terms of behavior the code above is only checking if a firewall `section` is still available for this security group and if yes, its skips it from removal on this cycle. Note that this is a safe guard for the concurrent events, deleting section and group at the same time. The real check whether the groups is used in a section or remote rule is done on the upper level in the realization function that makes this check against the Neutron DB. If we hae bug here it is probably in the DB call checking whether the group is used or not.

Also note that this might be caused by the legacy clean up code bellow, it does not make this check, and probably that is the problem.

```
            # TODO - remove after legacy provider is not supported
            try:
                self.legacy_provider.sg_members_realize({"id": os_id}, delete=True)
            except Exception:
                pass
```

PR #65, opened 2021-05-11
Development/context aware security ussuri

tsimchev:


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
```diff
@@ -81,8 +85,12 @@ def all(self, dryrun=False):
                 LOG.info("Dryrun:%s. Metadata refresh completed.", dryrun)
                 return
 
-            outdated = list(itertools.islice(port_outdated, slice))
-            slice -= len(outdated)
+            ## Don't count ports into synchronization limit, since ports
+            ## are not created by the agent they could exhaust the worker queue
+            ## and cause the agent to be stuck.
+            # outdated = list(itertools.islice(port_outdated, slice))
```
BarbarossaTM, line 91: Any particular reason to keep the old code here (instead of GIT history)?

PR #66, opened 2021-05-27
Added linter

fwiesel:
Unfortunately, the linter only runs with python3.5 (or later).


fwiesel:
Let's see, if the GitHub action even works.

PR #78, opened 2021-11-08
Ignore exceptions on double group creation

fwiesel:
If we try to create a group, which already exists, the api
throws an exception, but that isn't a problem for us, so
we can just continue as if we have created it

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
@@ -11,6 +11,7 @@ from networking_nsxv3.common.locking import LockManager
 from networking_nsxv3.plugins.ml2.drivers.nsxv3.agent import provider_nsx_mgmt
 from networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.constants_nsx import *
 from networking_nsxv3.prometheus import exporter
+import json
 
 LOG = logging.getLogger(__name__)
 
@@ -315,10 +316,10 @@ class Provider(provider_nsx_mgmt.Provider):
             return self.client.put(path=path, data=data).json()
         except Exception as e:
             with excutils.save_and_reraise_exception() as ctxt:
-                if 'already exists' in e.message:
+                if 'already exists' in e.args[1]:
                     ctxt.reraise = False
                     return self.client.patch(path=path, data=data).json()
-                return id
+                ctxt.reraise = True
     
     def _delete_sg_provider_rule_remote_prefix(self, id):
         self.client.delete(path=API.GROUP.format(id))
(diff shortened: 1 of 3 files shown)


joroaf:
Tested and verified.

joroaf (approved):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -308,8 +310,16 @@ def metadata(self, resource_type, os_id) -> provider_nsx_mgmt.ResourceMeta:
...
+        path = API.GROUP.format(id)
+        data = self.payload.sg_rule_remote(cidr)
+        try:
+            return self.client.put(path=path, data=data).json()
+        except Exception as e:
+            with excutils.save_and_reraise_exception() as ctxt:
+                if 'already exists' in e.args[1]:
+                    ctxt.reraise = False
+                    return self.client.patch(path=path, data=data).json()
+                ctxt.reraise = True
```
fwiesel, line 322: Setting ` ctxt.reraise = True` is redundant, as it is the default.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -308,8 +310,16 @@ def metadata(self, resource_type, os_id) -> provider_nsx_mgmt.ResourceMeta:
 
     def _create_sg_provider_rule_remote_prefix(self, cidr):
         id = re.sub(r"\.|:|\/", "-", cidr)
-        return self.client.put(path=API.GROUP.format(id),
-                               data=self.payload.sg_rule_remote(cidr)).json()
+        path = API.GROUP.format(id)
+        data = self.payload.sg_rule_remote(cidr)
+        try:
+            return self.client.put(path=path, data=data).json()
+        except Exception as e:
```
fwiesel, line 317: Can you narrow the exception a bit down?

tox.ini
```diff
@@ -39,6 +39,11 @@ basepython = python3.6
 commands =
     env TEST_OSPROFILER=1 stestr run 'networking_nsxv3.tests.unit'
 
+[testenv:u1]
```
fwiesel, line 42: I think, that should probably not be part of the commit.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -4,12 +4,14 @@
...
 from oslo_config import cfg
 from oslo_log import log as logging
+from oslo_utils import excutils
 
 from networking_nsxv3.common.constants import *
 from networking_nsxv3.common.locking import LockManager
 from networking_nsxv3.plugins.ml2.drivers.nsxv3.agent import provider_nsx_mgmt
 from networking_nsxv3.plugins.ml2.drivers.nsxv3.agent.constants_nsx import *
 from networking_nsxv3.prometheus import exporter
+import json
```
fwiesel, line 14: That import is not needed

PR #99, opened 2022-09-19
Don't process segment ports before migration

notandy:
This patch removes segment ports from the realization queue if
migration is not ongoing and the ports are existing in the management mode.

Fixes realization queue exhaustion caused by infinite segment-port realization


joroaf (commented):
I'll add also the reverse case when the mp_to_policy is enabled:
```
else:
                port_outdated = port_outdated.difference(seg_port_current)
                qos_outdated = qos_outdated.difference(seg_qos_current)
```

joroaf:
Actually we always have to filter-out the policy ports from manager meta and vice versa. I'll suggest this fix:
```
 seg_port_outdated = seg_port_outdated.difference(port_current)
 seg_qos_outdated = seg_qos_outdated.difference(qos_current)
 port_outdated = port_outdated.difference(seg_port_current)
 qos_outdated = qos_outdated.difference(seg_qos_current)
```

joroaf (approved):

PR #104, opened 2022-12-20
Fix/trunking port creation

joroaf:


sven-rosenzweig:
@joroaf I tested your feature. 

Test case 1: 
- create network & trunk (without subport)
- boot server with parent port
- attatch subport to trunk
--> Logical Port is created, openstack shows trunk in status ACTIVE


Test case 2: 
- create network & trunk (with subport)
-  boot server with parent port
--> Logical Port is crated, openstack shows trunk in status DOWN
--> Not sure why this is the case, but this still seems like a bug to me. 

Also I am not sure if we really need the method get_subport since it almost has the same functionality as the get_port method.

PR #108, opened 2023-03-13
Fix trunk RPC handlers

velp:
In these two commits:
https://github.com/sapcc/neutron/commit/129b823a8b610284ff95cc5e7b2f7b6ed3540172
https://github.com/sapcc/neutron/commit/69ef824069c3b1435a50c0d937ad45a23ff07c1b
the RPC was changed to use DBEventPayload from the neutron library instead of custom objects.

The new payload has different formats for different event types, here documentation: https://docs.openstack.org/neutron-lib/latest/contributor/callbacks.html#event-objects-dbeventpayload

In this PR we change our driver to have the correct handlers for RPC events coming from Neutron server.


notandy (approved):
lgtm

PR #117, opened 2023-07-10
Introduce unit test for metadata management

sven-rosenzweig:
The nsxt agent uses a method called is_manged to check wether an object (e.g. port, security group) is managed by itself or not. There has been an outage were a change to the 'is_managed' stopped the managemnt of objects created by a user called 'system'. A new unit test should cover if objects created by the 'system' user are still managed by agent. Those objects created by the system user are rather old. Newer objects are created by a user defined (currently 'admin') in the deployment of the agent.


notandy (approved):
lgtm

PR #123, opened 2023-08-17
Log all results which are not unique

sven-rosenzweig:
Elements expected to be unique (e.g. ports) not fulfilling this probably require some fixture. In order to make them more visible (sentry) the loglevel is increased to error.


notandy:
Hrm, isn't this change logging everything as "ambiguous?

notandy (changes requested):
Hrm, isn't this change logging everything as "ambiguous?

notandy:
> @notandy Also increased the log level to error in order to make them visible to us. In the past we had issues with duplicate entries. Do you agree on that?

I am fine with the log level increase, the PR doesn't not really changes anything else, right?

joroaf (approved):

notandy (approved):

PR #129, opened 2024-01-23
 NSXT Bulk API for realizing SGs

sven-rosenzweig:
For ports with more than 27 SGs, the approach of adding them as static member one by one results in a rpc-timout. Thus ports rely on the update_sg_group method executed after a port binding.

With this commit, we use the nsxt bulk API to add the staticmembership of ports directly after binding the port.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
@@ -217,16 +217,7 @@ class Payload(provider_nsx_mgmt.Payload):
                     "id": target_obj.get("id"),
                     "target_type": target_obj.get("resource_type"),
                     "children": [
-                        {
-                            o.get("resource_type"): {
-                                "path": o.get("path"),
-                                "parent_path": o.get("parent_path"),
-                                "id": o.get("id"),
-                                "resource_type": o.get("resource_type")
-                            },
-                            "resource_type": "Child{}".format(o.get("resource_type")),
-                            "marked_for_delete": o.get("marked_for_delete")
-                        } for o in child_objs
+                        { o } for o in child_objs
                     ]
                 }
             ]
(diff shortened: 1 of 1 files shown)


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -217,16 +217,7 @@ def infra(self, target_obj: dict, child_objs: List[dict]) -> dict:
...
-                            o.get("resource_type"): {
-                                "path": o.get("path"),
-                                "parent_path": o.get("parent_path"),
-                                "id": o.get("id"),
-                                "resource_type": o.get("resource_type")
-                            },
-                            "resource_type": "Child{}".format(o.get("resource_type")),
-                            "marked_for_delete": o.get("marked_for_delete")
-                        } for o in child_objs
+                        { o } for o in child_objs
```
joroaf, line 220: This simplification improvement does not work for the delete segment port scenario. Also, this is the reason why one of the functional tests failed. Suggest reviewing it to the old structure and adding the additional fields you need (if any). Also, it will be more readable.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -618,6 +617,33 @@ def _wait_to_realize(self, resource_type, os_id):
         exporter.REALIZED.labels(resource_type, status).inc()
         raise Exception("{} ID: {} did not get realized for {}s", resource_type, os_id, until * pause)
 
+    def _realize_bulk(self, resource_type: str, data, nsxt_res_type, dsl):
+        report = "Bulk Operation: Resource: {} with ID: {} is going to be %s.".format(resource_type, lambda: data[0]["id"])
+
+        LOG.info(report, "created")
+        path = "{}".format(API.INFRA)
+        
+        res = self.client.patch(path=path, data=data)
```
joroaf, line 626: We need a lock before the execution of the actual update/create operation. The challenge here is not to lock on the entire "SG_MEMBERS" type but for each group ID individually "member-{}".format(os_id). A semaphore implementation would be required.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -618,6 +617,33 @@ def _wait_to_realize(self, resource_type, os_id):
...
+        res.raise_for_status()
+
+        nsxt_objects = [child[nsxt_res_type] for child in data["children"][0]["children"]]
+        nsxt_objects = sorted(nsxt_objects, key=lambda x: x["id"])
+
+        # Fetch _revision data (needed to run put requests) for all updated objects
+        # This is needed to run put requests on objects since put relies on the correct _revision number
+        updated_nsxt_objs = self.client.get_all(path=API.SEARCH_DSL, params=API.SEARCH_DSL_QUERY(nsxt_res_type, dsl))
```
joroaf, line 634: I'm not sure about the search API approach. Ideally, we need to use GET /infra as we use the same endpoint for the update/create operation.

joroaf (changes requested):

PR #134, opened 2024-02-15
Deduplicate allowed address pairs

sven-rosenzweig:
NSX-T does not handle duplicates in manual address bindings. Causing the port binding to fail, deduplication needs to happen in the driver code. Duplication can happen if the OpenStack api user is adding the IP + MAC pair of a port also as allowed addresspair


sebageek (approved):
I like it

PR #136, opened 2024-03-26
Add IPv6 address of metadata service to sg rules

sven-rosenzweig:
So far only DHCPv6 and ICMPv6 traffic is refelcted in the default infrastructure rules of NSX-T. The IPv6 address of the metadata service requires clearing as well.

In order to reflect those changes in NSX-T, it requires a check if the rule configuration has changed. So far only creation of missing policies is supported.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/constants_nsx.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/constants_nsx.py
@@ -1,4 +1,4 @@
-
+import neutron_lib.constants as neutron_constants
 
 # IP_PROTOCOL_NUMBERS source
 # https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
@@ -233,7 +233,7 @@ DEFAULT_INFRASTRUCTURE_POLICIES = [
                 "id": "HTTP",
                 "display_name": "HTTP",
                 "source_groups": ["ANY"],
-                "destination_groups": ["169.254.169.254"],
+                "destination_groups": [neutron_constants.METADATA_V4_IP, neutron_constants.METADATA_V6_IP],
                 "services": ["/infra/services/HTTP"],
                 "service_entries": [],
                 "profiles": ["ANY"],
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
@@ -10,6 +10,7 @@ import time
 import netaddr
 import functools
 from oslo_cache import core as cache
+from dictdiffer import diff
 from oslo_config import cfg
 from oslo_log import log as logging
 from oslo_utils import excutils
@@ -113,6 +114,9 @@ class API(object):
 
     INFRA = "/policy/api/v1/infra"
 
+class Ignore(set):
+    def __contains__(self, key):
+        return set.__contains__(self, key[1])
 
 class PolicyResourceMeta(base.ResourceMeta):
     def __init__(self, id, unique_id, rev, age, revision, last_modified_time,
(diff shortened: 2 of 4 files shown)


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/constants_nsx.py
```diff
@@ -233,7 +233,7 @@
                 "id": "HTTP",
                 "display_name": "HTTP",
                 "source_groups": ["ANY"],
-                "destination_groups": ["169.254.169.254"],
+                "destination_groups": ["169.254.169.254", "fe80::a9fe:a9fe"],
```
sebageek: If you like constants we have `neutron_lib.constants.METADATA_V6_IP`. There's also one for v4.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -635,13 +635,27 @@ def _setup_default_infrastructure_rules(self):
             path = API.POLICY.format(policy["id"])
             res = self.client.get(path=path)
             if res.ok:
-                continue
+                if self._check_for_infrastructure_changes():
```
sebageek: What is this method? I haven't found it in the source. Do you want to call `_check_infrastructure_rules_for_updates()` here?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -635,13 +635,27 @@ def _setup_default_infrastructure_rules(self):
...
+                    self.client.put(path=path, data=policy).raise_for_status()
             elif res.status_code == 404:
                 LOG.info("Infrastructure Policy %s not found, creating...", policy["display_name"])
                 self.client.put(path=path, data=policy).raise_for_status()
             else:
                 res.raise_for_status()
 
+    def _check_infrastructure_rules_for_updates(self, policies_from_cfg, security_policies):
+        for rule_from_cfg in policies_from_cfg['rules']:
+            for rule_policy in sg_policy["rules"]:
```
sebageek: What is `sg_policy`? is that `security_policies` but with a different name?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -635,13 +635,27 @@ def _setup_default_infrastructure_rules(self):
...
+                if rule_policy['id'] == rule_from_cfg['id']:
+                    shared_keys = set(rule.keys()).intersection(set(rule_policy.keys()))
```
sebageek: [optional] I like using operators for set operations, but it depends of what you like yourself. For intersection you have the option of using `&`, so `set(rule) & set(rule_policy)` (the `.keys()` is implicit for dict-to-list-conversion but that's also a choice on how verbose you like to be, makes it more ovious that you're dealing with a dict).

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -635,13 +635,27 @@ def _setup_default_infrastructure_rules(self):
...
+                    diff = set(o for o in shared_keys if rule[o] != rule_policy[o])
+                    LOG.info("Infrastructure rule %s has changed in %s - new rule %s" % (rule_from_cfg, diff, rule_from_cfg))
```
sebageek: I'd pass the format attributes as arguments to `LOG.info()` instead of doing the formatting beforehand.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -635,13 +635,27 @@ def _setup_default_infrastructure_rules(self):
...
+                    return True
+                else:
+                    LOG.info("New infrastructure rule found %s" % rule_from_cfg)
```
sebageek: Same here for the Log args.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -635,13 +635,27 @@ def _setup_default_infrastructure_rules(self):
             path = API.POLICY.format(policy["id"])
             res = self.client.get(path=path)
             if res.ok:
-                continue
+                if self._check_for_infrastructure_changes():
+                    self.client.put(path=path, data=policy).raise_for_status()
```
joroaf: Should we update the default infra policy every time with the desired rules? What if the operator added his own rules? Is this the desired behaviour?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -113,6 +114,9 @@ class API(object):
 
     INFRA = "/policy/api/v1/infra"
 
+class Ignore(set):
```
sebageek, line 117: [optional] Could get another name that denotes more what it does. `IgnoreGivenKeys`? `DiffIgnoreKeys`? You also could move it closer to where it's used scope wise, on the other hand maybe you'll need it some day and then you want it in that package directly accessible? Dunno. Both options make sense to me.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -113,6 +114,9 @@ class API(object):
...
+    def __contains__(self, key):
+        return set.__contains__(self, key[1])
```
sebageek, line 119: would `super().__contains__(key[1])` work as well?

PR #137, opened 2024-04-09
Log execption in catch-all exception handers

sebageek:
For better debuggability we now log the full exception in catch-all exception handlers (i.e. where we don't know what the exception is about), as LOG.error only reports a single error log line, while LOG.exception gives us the full traceback. This also makes investigation easier with sentry, as we can see the local variables of each stack frame there.


sven-rosenzweig (approved):
Agree with you.

PR #141, opened 2024-10-11
Register DEFAULT group on agent start

sven-rosenzweig:
From upstream commit 4d3a2747 fixing bug (#1968606), registration of the group DEFAULT does not happen automatically anymore.

If not loaded, agent fails with 'oslo_config.cfg.NoSuchOptError: no such option base_mac in group [DEFAULT]


.github/workflows/coverage.yaml
joroaf, line 1: Don't you need coverage tests anymore?

sven-rosenzweig: force-pushed, unwanted changes are gone.

PR #142, opened 2024-11-19
[Caracal] Migrate github workflow and Fix Tests

sven-rosenzweig:
Change python version and basebranch according to the Caracal release to
version 3.10 and 2024.1-m3.

Loosen constraints for the installation of neutron and
neutron-lib in the test-requirements.txt as the upperbound was set to
Neutron Yoga release. Within the tox configuration request the
installation of yoga from the current stable/2024.1-m3 branch.

For the DB unit tests remove the subnetpool as this is not needed for
a subnet creation. Without the subnetpool, the db unit test test_port_sg
fails as the result of a DB query returns the cartesian product between
the tables SecurityGroupPortBindings and StandardAttributes. Thus the
length of the result set is lowered from 9 to 8.
This is a minor bug which will be fixed in a separate PR.

### DIFF
--- a/networking_nsxv3/db/db.py
+++ b/networking_nsxv3/db/db.py
@@ -166,7 +166,8 @@ def get_port(context, host, port_id):
         PortBinding.vif_details,
         PortBinding.status
     ).join(
-        StandardAttribute,
+        StandardAttribute
+    ).join(
         PortBinding
     ).filter(
         Port.id == port_id,
--- a/networking_nsxv3/db/db.py
+++ b/networking_nsxv3/db/db.py
@@ -13,11 +13,13 @@ from neutron.db.qos.models import (QosBandwidthLimitRule, QosDscpMarkingRule,
 from neutron.plugins.ml2.models import PortBinding, PortBindingLevel
 from neutron.services.trunk import models as trunk_model
 from neutron_lib.api.definitions import portbindings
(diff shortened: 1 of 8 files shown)


notandy (approved):

requirements.txt
```diff
@@ -5,7 +5,7 @@
 pbr>=1.6
 Babel>=1.3
 
-neutron-lib>=1.13.0             # Apache-2.0
+neutron-lib                     # Apache-2.0
```
mutax: Thats the only one without min. version but the test requirements have `1.13.0` - intentionally?

networking_nsxv3/db/db.py
```diff
@@ -381,10 +403,10 @@ def get_security_group_port_ids(context, host, security_group_id):
         "GROUP BY port_id"
     )
 
-    ports_with_sg_count = ses.execute(sql)
-    return ports_with_sg_count
+    return [row._mapping for row in ses.execute(sql)]
```
mutax, line 406: is this a compatible change?

mutax (approved):

requirements.txt
sven-rosenzweig: good catch, with my last rebase I wanted to add `>=1.13.0 ` again

mutax (approved):

PR #143, opened 2025-02-07
Fix Attribute Error when Metadata is Missing

sven-rosenzweig:
In some cases, metadata information (including age) may be missing for an OpenStack ID. Previously, this caused an exception, leading to an aborted sync.

Since the sync is a preventive measure to synchronize objects of a certain age — most likely unchanged from OpenStack's perspective — this commit ensures the process continues even when metadata is absent.

To improve debugging, we now log the resource_type and os_id when metadata is missing.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
@@ -1136,7 +1136,14 @@ class Provider(base.Provider):
         return outdated, current
 
     def age(self, resource_type: str, os_ids: List[str]):
-        return [(resource_type, id, self.metadata(resource_type, id).age) for id in os_ids]
+        aged_resources = []
+        for id in os_ids:
+            meta_data = self.metadata(resource_type, id)
+            if meta_data:
+                aged_resources.append((resource_type, id, meta_data.age))
+            else:
+                LOG.info("Resource: %s with ID: %s has no metadata", resource_type, id)
+        return aged_resources
 
     # overrides
     def sanitize(self, slice):


sven-rosenzweig:
With this commit we should get rid of the following [exception](https://sentry.qa-de-1.cloud.sap/share/issue/405a535f54c141bf8ada3f9508d9cd4b/).

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -1136,7 +1136,14 @@ def outdated(self, resource_type: str, os_meta: Dict[str, dict]):
...
 
     def age(self, resource_type: str, os_ids: List[str]):
-        return [(resource_type, id, self.metadata(resource_type, id).age) for id in os_ids]
+        aged_resources = []
+        for id in os_ids:
+            meta_data = self.metadata(resource_type, id)
+            if meta_data:
+                aged_resources.append((resource_type, id, meta_data.age))
+            else:
+                LOG.info("Resource: %s with ID: %s has no metadata", resource_type, id)
```
sebageek, line 1145: [optional] Would write `"Resource %s with ID %s has no metadata"` or `"Object %s: Missing metadata for resource %s"` - the two `:` seem weird to me.

sebageek (approved):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
mutax: it follows the [logging format used here](https://github.com/sapcc/networking-nsx-t/blob/ae2eaec572f9106623bf6a4891544dc3aa0caf42/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py#L39
)  in the same file -- would stick with it so searching the logs can use the same expression to match

notandy (approved):

PR #144, opened 2025-02-10
Prevent Jobs from Starving in the Active Queue

sven-rosenzweig:
The active queue uses a heap-based priority queue, sorting jobs by priority.
Tasks received via RPC from Neutron have the highest priority.
Jobs from the sync loop have lower priority.
Jobs moved from the passive to the active queue retain their original priority.

Every interval x, the agent triggers a background sync, placing objects that
need synchronization into the passive queue.
These jobs may be moved to the active queue if capacity allows.

However, when the agent is fully occupied (handling 40 tasks simultaneously),
and more high-priority jobs are coming in, lower-priority jobs in the active
queue may become stuck for some time, as higher-priority
jobs are continuously processed first.

With our approach sort stability is not given.
Adding new jobs with HIGHEST priority does not guarantee they are
returned in the order the have been added. Details can be found in
the heapq documentation (https://docs.python.org/3/library/heapq.html#priority-queue-implementation-notes).

To overcome this potential drawbacks the active queue becomes a FiFo
Queue. Jobs on the active queue are treated with same equality and are
processed with insertion order.

### DIFF
--- a/networking_nsxv3/common/synchronization.py
+++ b/networking_nsxv3/common/synchronization.py
@@ -90,12 +90,51 @@ class Runnable(object):
         return (not self.__eq__(other))
 
     def __lt__(self, other):
-        """ Order Runnable by their priority """
+        """ Order Runnable by their priority
+        Only the passive queue is ordered by priority.
+        The active queue is FiFo.
+        """
         return self.priority < other.priority
 
     def __hash__(self):
         return hash(self.__repr__())
 
+class UniqFiFoQueue(eventlet.queue.Queue):
+    """
+    A subclass of :class:`Queue` that maintains job order by insertion.
+
+    Problem with the old approach:
+    Jobs may starve in the active queue.
+
+    - The internal sync run adds up to 20 "outdated" objects to the passive queue.
+    - If space is available, jobs move from the passive to the active queue.
+    - The active queue prioritizes jobs based on priority while enforcing uniqueness
+      (determined by OpenStack ID and execution method).
+
+    Issues:
+    1. Job starvation – If high-priority jobs keep arriving and the agent is at full capacity,
+       low-priority jobs may never get processed.
+    2. Blocking of new high-priority jobs – Lower-priority jobs in the passive queue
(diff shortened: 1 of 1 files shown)


networking_nsxv3/common/synchronization.py
```diff
@@ -90,12 +90,35 @@ def __ne__(self, other):
...
     def __hash__(self):
         return hash(self.__repr__())
 
+class UniqFiFoQueue(eventlet.queue.Queue):
+    """A subclass of :class:`Queue` that retrieves the oldest added entries first.
+        Use a FiFo queue to store the Active JOBS.
+        Before this sync Loop was a priority queue. Every sync run at most 20 "outdated"
+        NSX-T objects are added to a passive queue. If the active queue is empty, the jobs gets copied to the active queue.
+        The active queue is ordered by the priority of the jobs.
+        Changing this to FiFo will make sure not job starve in the acitve queue as the priority is to low.
```
mutax: Would add a comment here to emphasize the other problem with the queue. 
The problem with starvation is amplified by the fact, that items from the passive queue with lower priority will prevent new jobs being added with higher priority -- see the` __eq__` method of `Runnable ` in line 78 and the comment. 
And maybe add a comment that we suspect the order to be important, e.g. first create the ports, then the SGs.

mutax (dismissed):

networking_nsxv3/common/synchronization.py
```diff
@@ -90,12 +90,35 @@ def __ne__(self, other):
...
+        Also the order of jobs with HIGEST priority will be kept.
+    """
+
+    def _init(self, maxsize):
+        self.queue = collections.deque()
+
+    def _put(self, item):
+        # Add item to the right side of the deque
+        self.queue.append(item)
```
mutax: on a second look - the uniqueness was dropped here :)

sven-rosenzweig: Nice catch. Thanks for that.

mutax (approved):

mutax:
code cherry picked into PR #145

PR #145, opened 2025-02-14
Fix blocking worker threads, possible race conditions and improve logging

mutax:
With this PR we modify the scheduling of jobs to the worker pool, improve logging and allow single unit tests to be executed.

On the way we fixed several bugs and inconsistencies in the existing code, required to implement the changes.

Please refer to the commit message of the individual changes, the explanation for the changes in the scheduling is rather long.

### DIFF
--- a/networking_nsxv3/common/synchronization.py
+++ b/networking_nsxv3/common/synchronization.py
@@ -1,10 +1,13 @@
 """
 Synchronization - classes related concurrent execution scheduling and limits
 """
+from typing import Callable, Union, Optional, List
+
 import eventlet
 eventlet.monkey_patch()
 
 import networking_nsxv3.prometheus.exporter as EXPORTER
+from networking_nsxv3.common.locking import LockManager
 from oslo_log import log as logging
 from oslo_config import cfg
 import enum
@@ -65,36 +68,178 @@ class Identifier(object):
         return False
 
 
+CBPARAMS = Union[str, dict]
+
+
 class Runnable(object):
 
-    def __init__(self, idn, fn, priority=Priority.LOWEST):
+    def __init__(self, fnparams: CBPARAMS, fn: Callable[[CBPARAMS], None], priority=Priority.LOWEST):
         self.priority = priority
-        self.idn = idn
-        self.fn = fn
+        self._fnparams = fnparams
+        self._fn = fn
+
+        # contradicting to code comments, we sometimes get a dictionary
+        # as parameter for the callback. Some data from that
+        # dictionary is then actually used, e.g. revision and resource id.
+        # Not going to happen. Fixing that would require major
+        # refactoring of the rpc and realizer objects. Not going to happen.
+        #
+        # Instead, in these cases we will still use the openstack id and the
+        # name of the callback function to prevent parallel running of jobs.
+        # But when a job is submitted to the Rerunner, it will keep track of the
+        # dictionary contents if one exists and keep these for rerunning. So essentially
+        # we create subjobs.
+        # Realistically we need a better data structure than the active queue and Rerunner.
+
+        # fnparams is a str for most of the callbacks:
+        if isinstance(fnparams, str):
+            self.idn = fnparams
+        elif isinstance(fnparams, dict):
+            self.idn = fnparams['id']
+        else:
+            self.idn = str(fnparams)
+            LOG.warning('unexpected type %s for job parameters of %s', type(fnparams), fn.__name__)
+
+        self._runcount = 0
+        self._created = time.time()
+        self._scheduled = None
+        self._started = None
+        self._jobdone = None
+        self._rescheduled = None
+
+    @property
+    def identifier(self) -> tuple:
+        return self.idn, self._fn.__name__
+
+    def debugid(self)->str:
+        return str((self.idn, self._fn.__name__, str(self._fnparams)))
+
+    def set_scheduled(self):
+        """ called when we submit the job to the worker pool """
+        self._scheduled = time.time()
+        self._started = None
+        self._jobdone = None
+        self._rescheduled = None
+
+    def _set_start(self):
+        """ called in our wrapper when we actually start fn """
+        # we need to reset the other timings
+        # because we use the same job for rerunning
+        self._started = time.time()
+        self._jobdone = None
+        self._rescheduled = None
+        self._runcount += 1
+
+    def _set_done(self):
+        """ called in our wrapper when fn returns """
+        self._jobdone = time.time()
+        self._rescheduled = None
+
+    def set_rescheduled(self):
+        """ called when the job is taken out of the rerun-queue """
+        self._rescheduled = time.time()
+
+    def get_statline(self) -> str:
+
+        age = f"{time.time() - self._created:0.4f}"
+
+        scheduled = '-'
+        started = '-'
+        runtime = '-'
+        rescheduled = '-'
+
+        if self._scheduled and self._created:
+            scheduled = f"{self._scheduled - self._created:0.4f}"
+
+        if self._started and self._scheduled:
+            started = f"{self._started - self._scheduled:0.4f}"
+
+        if self._jobdone and self._started:
+            runtime = f"{self._jobdone - self._started:0.4f}"
+
+        if self._rescheduled and self._jobdone:
+            rescheduled = f"{self._rescheduled - self._jobdone:0.4f}"
+
+        return (f"timings for job {self} - runcount: {self._runcount} age: {age} "
+                f"scheduled: {scheduled} started: {started} runtime: {runtime} rescheduled: {rescheduled}")
+
+    def execute(self):
+        self._set_start()
+        try:
+            self._fn(self._fnparams)
+        finally:
+            self._set_done()
 
     def __repr__(self):
-        return str(self.idn)
+        # lets not just use the object id, maybe
+        return str(self.identifier)
 
     def __eq__(self, other):
         """
         Note, the priority is not part of the comparison
         Thus if a runnable with higher priority is about to be
         added to the queue it will be rejected silently.
+        To prevent starvation, the queue will update the priority of the
+        existing element, in case it was lower than the item that was about to be added.
         """
         if isinstance(other, Runnable):
-            return (self.idn == other.idn and self.fn == other.fn)
+            return self._fnparams  == other._fnparams  and self._fn == other._fn
         else:
             return False
 
     def __ne__(self, other):
-        return (not self.__eq__(other))
+        return not self.__eq__(other)
 
     def __lt__(self, other):
-        """ Order Runnable by their priority """
+        """ Order Runnable by their priority
+        Only the passive queue is ordered by priority.
+        The active queue is FiFo.
+        """
+        # if the priority is equal, we want to order
+        # by creation time to handle oldest jobs first
+        if self.priority == other.priority:
+            return self._created < other._created # noqa
+
         return self.priority < other.priority
 
-    def __hash__(self):
-        return hash(self.__repr__())
+
+class UniqFiFoQueue(eventlet.queue.Queue):
+    """
+    A subclass of :class:`Queue` that maintains job order by insertion.
+
+    Problem with the old approach:
+    Jobs may starve in the active queue.
+
+    - The internal sync run adds up to 20 "outdated" objects to the passive queue.
+    - If space is available, jobs move from the passive to the active queue.
+    - The active queue prioritizes jobs based on priority while enforcing uniqueness
+      (determined by OpenStack ID and execution method).
+
+    Issues:
+    1. Job starvation – If high-priority jobs keep arriving and the agent is at full capacity,
+       low-priority jobs may never get processed.
+    2. Blocking of new high-priority jobs – Lower-priority jobs in the passive queue
+       can prevent new high-priority jobs from being added due to the uniqueness constraint
+       (see the `__eq__` method of the `Runnable` class).
+    3. Adding a job with the **HIGHEST** priority does not
+       guarantee execution in insertion order. See
+       https://docs.python.org/3/library/heapq.html#priority-queue-implementation-notes
+       for details.
+    """
+
+    def _init(self, maxsize):
+        self.queue = collections.deque()
+
+    def _put(self, item):
+        if item not in self.queue:
+            # Add item to the right side of the deque
+            self.queue.append(item)
+            self._put_bookkeeping()
+        else:
+            LOG.info("Not adding item %s to fifo queue, already present!", item)
+
+    def _get(self):
+        return self.queue.popleft()
 
 
 class UniqPriorityQueue(eventlet.queue.Queue):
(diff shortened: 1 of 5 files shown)


networking_nsxv3/common/synchronization.py
```diff
@@ -97,7 +102,8 @@ def __lt__(self, other):
         return self.priority < other.priority
 
     def __hash__(self):
-        return hash(self.__repr__())
+        # with the original repr this is broken, as __eq__ takes the fn into account as well!
+        return hash((self.idn, self.fn))
```
sven-rosenzweig: Can we use ``self.identifier`` here? So `self.fn.__name__` instead of `self.fn`.

sven-rosenzweig: Do we need to override ``__hahs__``? 
According to the docs: 
Called by built-in function [hash()](https://docs.python.org/2/library/functions.html#hash) and for operations on members of hashed collections including [set](https://docs.python.org/2/library/stdtypes.html#set), [frozenset](https://docs.python.org/2/library/stdtypes.html#frozenset), and [dict](https://docs.python.org/2/library/stdtypes.html#dict). [__hash__()](https://docs.python.org/2/reference/datamodel.html#object.__hash__) should return an integer

We simply add Runnable in a Queue (managing a list), right?

networking_nsxv3/common/synchronization.py
```diff
@@ -149,15 +149,29 @@ class UniqPriorityQueue(eventlet.queue.Queue):
...
-        if item not in self.queue:
-            heappush(self.queue, item)
-            self._put_bookkeeping()
-        else:
-            LOG.info("Not adding item %s to prio queue, already present!", item)
+    def _put(self, item):
 
-    def _get(self, heappop=heapq.heappop):
-        return heappop(self.queue)
+        try:
```
sven-rosenzweig, line 252: Do we really need this? 
As the passive queue is the only UniqPriorityQueue left. As for the current implementation all jobs added to the passive queue have the same priority. 

The active queue becomes a FiFo Queue, priority won't matter anymore.

networking_nsxv3/common/synchronization.py
```diff
@@ -173,6 +174,93 @@ def _put(self, item):
     def _get(self):
         return heapq.heappop(self.queue)
 
+class JobRerunner():
+    """ Threadsave data structure to reschedule jobs when already running
```
sven-rosenzweig: Good explanation. 
Maybe we could emphasise, that a job is runs at max twice as long as it is running. So it does not matter if we increase the counter to 100 or 2.

networking_nsxv3/common/synchronization.py
```diff
@@ -173,6 +174,93 @@ def _put(self, item):
...
+                job = None
+        if job:
+            LOG.info("Found rerunnable job: %s", job)
+        else:
+            LOG.debug("No job to rerun in queue")
+        return job
+
+    def job_done(self, job: Runnable):
+        LOG.debug("done job %s", job)
+        with LockManager.get_lock("JobRerunner-running"):
```
sven-rosenzweig: Use a const here? it is also used in the ``add_job`` function.

networking_nsxv3/common/synchronization.py
```diff
@@ -232,10 +321,29 @@ def _start(self):
                 if self.active() < self._idle and self.passive() > 0:
                     self._active.put_nowait(self._passive.get_nowait())
                     self._passive.task_done()
-                job = self._active.get(block=True, timeout=TIMEOUT)
-                LOG.info(MESSAGE.format("Processing", job.idn, Priority(job.priority).name, job.fn.__name__))
-                self._workers.spawn(job.fn, job.idn)#.wait()
-                self._active.task_done()
+                job = self._rerunner.pop()
+                from_queue = False
```
sven-rosenzweig: Would rename it to ``pulled_from_active_queue``

networking_nsxv3/common/synchronization.py
```diff
@@ -173,6 +174,93 @@ def _put(self, item):
...
+    When the runner is done with a job, it will notify us and we can either
+    remove the id from our list or, if a new job was added in the meantime,
+    mark it as ready for re-execution.
+
+    Before taking a new job from the active queue, we first check this data
+    structure for a ready job.
+    """
+
+    def __init__(self):
+        self._ready = collections.deque()
```
sven-rosenzweig: this is the queue storing jobs which needs to be rerun? Would change the naming here.

sven-rosenzweig: Also the job ``def pop(self)`` is misleading.

networking_nsxv3/common/synchronization.py
mutax: `__hash__` was overwritten upstream, but buggy -- if two objects compare equal, they should also have identical hash values and vice-versa.

After fixing that I was able to use a ` job: Runnable` as keys in my dictionaries for the rerunner and stats.

Similarily we have to use `fn` and not just `fn.__name__` , because the name does not include the package name and in the `__eq__`  method fn is compared and not the __name__.

 Hence `__eq__` would be false, but hash() would return the same hash if a function from another python package is used, e.g. to overwrite the callback.

networking_nsxv3/common/synchronization.py
mutax: We should be able to use a UniqFifoQueue for the passive queue as well now, that is correct, because we handle the only 2 priorities used in two queues now. As long as we do not need more priorities that will work. 

While in principle I am open for that I would do that change separately in case we find out we would benefit from actually using more priorities to, e.g. , first delete ports and then create new ones/call update.

networking_nsxv3/common/synchronization.py
mutax: Currently there is one comment addressing that, should we make it more obvious?
 https://github.com/sapcc/networking-nsx-t/blob/12ae5a49f44baaf2f26ebbcb1c30876a3b7fbd80/networking_nsxv3/common/synchronization.py#L236
 _Note that a job will intentionally only be re-executed once, independently of how often it was added while a worker was already running it._

networking_nsxv3/common/synchronization.py
mutax: do we want to do that or keep it as is for now?

networking_nsxv3/common/synchronization.py
mutax: done, but not sure if that made the code more readable, though

networking_nsxv3/common/synchronization.py
mutax: I skipped active in the name because I found it too long.

networking_nsxv3/common/synchronization.py
mutax: renamed

networking_nsxv3/common/synchronization.py
```diff
@@ -72,11 +72,66 @@ def __init__(self, idn, fn, priority=Priority.LOWEST):
...
+        """ called when the job is taken out of the rerun-queue """
+        self._rescheduled = time.time()
+
+    def get_statline(self) -> str:
+
+        age = f"{time.time() - self._created:0.4f}"
+
+        scheduled = '-'
+        started = '-'
+        jobdone = '-'
```
sven-rosenzweig: I would ``jobdone`` to ``runtime``. the label in the format string says so: ``runtime: {jobdone}``.

networking_nsxv3/common/synchronization.py
```diff
@@ -222,10 +278,13 @@ def pop(self) -> Runnable:
         with LockManager.get_lock("JobRerunner-deque"):
             try:
                 job = self._ready.popleft()
+                job.set_rescheduled()
                 LOG.debug("JobRerunner had rerunnable job: %s", job)
```
sven-rosenzweig: Do we need this debugging log?

networking_nsxv3/common/synchronization.py
```diff
@@ -236,17 +295,21 @@ def job_done(self, job: Runnable):
             if count == 1:
                 del self._running[job]
                 LOG.debug("JobRerunner job %s is done, no reruns requested", job)
```
sven-rosenzweig: Do we need this debugging log?

networking_nsxv3/common/synchronization.py
```diff
@@ -260,25 +323,26 @@ def add_job(self, job: Runnable) -> bool:
...
+            LOG.debug("JobRerunner job %s already running, marked for rescheduling, count: %d ", job, count)
+
+            sum = 0
+            for job, scount in self._running.items():
+                sum += scount
+                LOG.debug("JobRerunner stat: job %s is running, submission count: %d", job, scount)
+
+        # let's log these as info for debugging, they should be sufficient in prod
+        # to find issues with the JobRerunner:
+        LOG.info("JobRerunner stat: %d jobs waiting, total submission count: %d", len(self._running), sum)
```
sven-rosenzweig: What insights do we get with the total count? 
Does it make sense to replace ``%d jobs waiting`` with ``%d jobs running``  as we take the infromation from ``len(self._running)``?

networking_nsxv3/common/synchronization.py
mutax: right, forgot to rename the var when I added started and the calls in the wrapper - will do!

networking_nsxv3/common/synchronization.py
mutax: We are currently running with INFO so it does not hurt in case we need to debug the changed scheduling -- would keep it until we are sure we don't need it.

networking_nsxv3/common/synchronization.py
mutax: We are currently running with INFO so it does not hurt in case we need to debug the changed scheduling -- would keep it until we are sure we don't need it.

networking_nsxv3/common/synchronization.py
mutax: The total count gives us a metric to compare to the 40 worker threads, as that gives us an indication how big the problem we are trying to fix (all workers blocked with the same requests) is in prod. If we grep the log and see this number increasing without the number of jobs increasing, we know that we would have blocked all workers with the current code.

networking_nsxv3/common/synchronization.py
sven-rosenzweig: But the same info is already contained in the info log (job is rerun, id and function name)?

networking_nsxv3/common/synchronization.py
mutax: jobs waiting was meant to say "jobs waiting for reexecution" but you are right, that is wrong as not all of them are waiting for that.  Jobs running should be more or less correct, although this is just the number of jobs tracked in this class and not directly coupled to the worker pool. I think I will rename it to 'jobs tracked' -- thats less confusing

sven-rosenzweig: I like 'jobs tracked'.

networking_nsxv3/common/synchronization.py
mutax: the ui didn't show the next line in the conversation pane, did not see the info below. I reworded the log messages of the info message to be consistent, also the statusline method uses self in the Runnable, so all log messages will include the same pattern in case we modify repr later. debug removed.

networking_nsxv3/common/synchronization.py
mutax: the ui didn't show the next line in the conversation pane, did not see the info below. I reworded the log messages of the info message to be consistent, also the statusline method uses self in the Runnable, so all log messages will include the same pattern in case we modify repr later. debug removed.

networking_nsxv3/common/synchronization.py
sven-rosenzweig: As discussed, we keep this for now because we might need it in future again.

sven-rosenzweig (approved):

networking_nsxv3/tests/unit/realization/test_coordination.py
```diff
@@ -0,0 +1,182 @@
...
+        job.set_rescheduled()
+        self.assertNotIn('rescheduled: -', job.get_statline())
+        self.assertNotIn('runtime: -', job.get_statline())
+
+        callback.assert_called()
+
+
+
+
+
```
felix-kaestner:
```suggestion
```

networking_nsxv3/common/synchronization.py
```diff
@@ -90,25 +151,204 @@ def __ne__(self, other):
...
+        return heapq.heappop(self.queue)
+
+
+class JobRerunner():
+    """ Thread save data structure to reschedule jobs when they are already running
+
+    When a job is retrieved from the active queue and already running in a worker thread,
+    there is a chance that another job for the same object is added to the active
+    queue and also started in a worker. While there then is some locking happening
+    that should prevent race conditions, this still blocks the worker thread(s),
```
joker-at-work: minor: "While there then is some locking happening" sounds strange

networking_nsxv3/common/synchronization.py
```diff
@@ -90,25 +151,204 @@ def __ne__(self, other):
...
+                # the job is still running or they get re-queued.
+                # if a job is in the ready deque it will at some point
+                # re-appear and so we can forget about the counter.
+                LOG.info("JobRerunner (done, %d reruns requested) %s", count - 1, job.get_statline())
+                del self._running[job]
+                with LockManager.get_lock(self._lname_torerun):
+                    self._to_rerun.append(job)
+            else:
+                # prevent the error from spreading
+                del self._running[job]
```
joker-at-work: This could raise `KeyError` if the `job`is not in there anymore. We can still get `count == 0`, because of `self._running.get()` in line 300.

networking_nsxv3/common/synchronization.py
```diff
@@ -90,25 +151,204 @@ def __ne__(self, other):
...
+
+        """
+        with LockManager.get_lock(self._lname_running):
+            count = self._running.get(job, 0)
+            if count <= 0:
+                # no job running, we can run the job
+                # if the counter is < 0 our accounting is wrong,
+                # so we fix it and run the job.
+                self._running[job] = 1
+                LOG.debug("JobRerunner no identical job is currently running, can start %s", job)
```
joker-at-work, line 555: Why no `job.get_statline()`?

networking_nsxv3/common/synchronization.py
```diff
@@ -90,25 +151,204 @@ def __ne__(self, other):
...
+                return True
+
+            self._running[job] = count + 1
+            LOG.debug("JobRerunner job %s already running, marked for rescheduling, count: %d ", job, count)
```
joker-at-work: Why no `job.get_statline()`?

networking_nsxv3/common/synchronization.py
```diff
@@ -90,25 +151,204 @@ def __ne__(self, other):
...
+
+            sum = 0
+            for job, scount in self._running.items():
+                sum += scount
+                LOG.debug("JobRerunner stat: job %s is running, submission count: %d", job, scount)
```
joker-at-work: Why no `job.get_statline()`?

joker-at-work: On every submission we print the stats? Don't you want to mark this as "Please remove once we're ready for prod"? Seems quite excessive. How large do you think `self._running`? will become?

networking_nsxv3/common/synchronization.py
```diff
@@ -169,10 +410,31 @@ def _start(self):
...
+                    LOG.info(MESSAGE.format("Processing", job.idn, Priority(job.priority).name, job.fn.__name__))
+
+                    # ideally we would be able to add a callback to the
+                    # greenthread, but this is hidden in the pool, so
+                    # let's wrap the function once more.
+                    def wrap(rerun, ajob):
+                        ajob.set_start()
+                        ajob.fn(ajob.idn)
+                        ajob.set_done()
+                        rerun.job_done(ajob)
```
joker-at-work: Why does this not need a try-finally?

joker-at-work (commented):
first glance over it

networking_nsxv3/common/synchronization.py
mutax: max 40, as we have 40 workers running. If all workers are busy, the loop adding new jobs will not add more jobs.

mutax:
marked PR as draft, need some more work -- contradicting the code and comments in the code, jid is not always a string containing an OS-ID, sometimes its a dictionary with the parameters for the actual change.  This of course breaks a lot of assumptions.

networking_nsxv3/common/synchronization.py
joker-at-work: So that could be 40 lines per ... second per BB? Every time we try to run a job hm ...

networking_nsxv3/common/synchronization.py
mutax: was added and the whole logic moved into the runnable, so the set methods are no longer public and the runnable is under control of whats happening.

networking_nsxv3/common/synchronization.py
mutax: rewritten.

networking_nsxv3/common/synchronization.py
mutax: the statusline is relatively verbose and the stats are only really interesting when a state change occurs, e.g. when a job is started or done or rescheduled. Its sufficient to get the identifier when debugging the flow of execution, I think.

networking_nsxv3/common/synchronization.py
mutax: Same as above, here we are only really interested in how often a job has been re-requested, so no need for all the details

networking_nsxv3/common/synchronization.py
mutax: see above

mutax: only when using debug, which we usually only do in qa - so I think in those cases its okay?

joker-at-work: Ah, I thought we'd be running at debug-level, but we're on INFO, you're right.

networking_nsxv3/common/synchronization.py
```diff
@@ -65,50 +68,494 @@ def retry_next(self):
...
+    def __init__(self):
+        self._job_identifier: Optional[str] = None
+        self._runnables: List[tuple[int, Runnable]] = []
+
+    def __len__(self):
+        return len(self._runnables)
+
+    @property
+    def size(self):
+        return sum( count for count, _ in self._runnables)
```
felix-kaestner:
```suggestion
        return sum(count for count, _ in self._runnables)
```

networking_nsxv3/common/synchronization.py
```diff
@@ -129,18 +576,24 @@ class Runner(object):
     def __init__(self, active_size=INFINITY, passive_size=INFINITY,
                  workers_size=1):
         # if queue_size is < 0, the queue size is infinite.
-        self._active = UniqPriorityQueue(maxsize=active_size)
+        self._active = UniqFiFoQueue(active_size)
```
felix-kaestner:
```suggestion
        self._active = UniqFiFoQueue(maxsize=active_size)
```

networking_nsxv3/common/synchronization.py
```diff
@@ -238,14 +273,179 @@ def _get(self):
         return heapq.heappop(self.queue)
 
 
+class JobList():
+    """ List of Runnables with the same identifier but different parameters
+
+        In this datastructure we are keeping track of the jobs and their parameters for the JobRerunner,
+        based on the identifier of the jib, which is telling us the type of job (callback name) and openstack id.
+
+        There can be multiple similar jobs (same id, same method) but with different parameters, then parameter
```
sven-rosenzweig, line 282: Dicts only for ``address groups`` and ``logging``. Could mention this in this docstring.

networking_nsxv3/common/synchronization.py
```diff
@@ -238,14 +273,179 @@ def _get(self):
...
+        When a job is done, we will look at our list, and decrease the counter.
+        If the counter is 0, the job was not submitted a second time, and we can remove
+        it from the list.
+        If the counter is not 0, the job was requested to run again, and we keep it in the
+        list with the updated counter.
+
+        done will then choose a job from the list, that is supposed to run again,
+        remove it from the list and return it to the JobRerunner.
+
+        This might be the same job that was just finished or it could be a different one.
```
sven-rosenzweig: With different one == same job but other parameters, right?

networking_nsxv3/common/synchronization.py
```diff
@@ -238,14 +273,179 @@ def _get(self):
...
+        This method adds those jobs to this list, but checks for those extra parameters.
+
+        See Class documentation for more details.
+
+        @return: True if the same job was not present yet, False if one already existed
+        """
+
+        if self._job_identifier is not None:
+            if job.identifier != self._job_identifier:
+                raise ValueError("Can only add jobs of same type to a JobList")
```
sven-rosenzweig, line 355: What happens if we raise this exception? Might a job not be added to the queue?

networking_nsxv3/common/synchronization.py
```diff
@@ -238,14 +273,179 @@ def _get(self):
...
+        else:
+            self._job_identifier = job.identifier
+
+        # search through our list and update the counter or append the job:
+        for index, (count, existing_job) in enumerate(self._runnables):
+            if job == existing_job:
+                count += 1
+                self._runnables[index] = (count, existing_job)
+                if count > 1:
```
sven-rosenzweig: Does not work as stated in the comment.

networking_nsxv3/common/synchronization.py
```diff
@@ -296,26 +496,35 @@ def get_rerunnable(self) -> Runnable:
 
     def job_done(self, job: Runnable):
         LOG.debug("JobRerunner job_done called for %s", job)
+
         with LockManager.get_lock(self._lname_running):
-            count = self._running.get(job, 0)
 
-            if count == 1:
-                del self._running[job]
+            joblist: JobList = self._running[job.identifier]
```
sven-rosenzweig, line 514:
```
            joblist: JobList = self._running[job.identifier]
            if joblist.empty():
                try:
                    del self._running[job.identifier]
                except KeyError:
                    # should never happen, we just fetched it above
                    pass
                LOG.info("JobRerunner (done, no reruns requested) %s", job.get_statline())
                return
            
            # TODO(mutax): make debug after nsx-t issues solved
            LOG.info("After job done %s", joblist)

            next_job = None
            try:
                next_job = joblist.done(job)
            except Exception as e:
                LOG.error("Error getting next job after a job was done: %s", e)

            if next_job: 
                LOG.info("JobRerunner (done, got rerun) done: %s next: %s", job.get_statline(), next_job.get_statline())
                with LockManager.get_lock(self._lname_torerun):
                    self._to_rerun.append(next_job)
            else:
              LOG.info("JobRerunner (done, no reruns requested) %s", job.get_statline())
```
just an alternative suggestion for this. not 100% sure if it covers your implementation.

networking_nsxv3/common/synchronization.py
mutax: I found these:  `address_group_update `and `{enable, disable, update}_policy_logging` in AgentRealizer,  are those all?

networking_nsxv3/common/synchronization.py
mutax: yes, will clarify

networking_nsxv3/common/synchronization.py
mutax: Yes, this is a programming error and should never happen -- there is not much we can do at this point than raise an exception and log it as an error.

networking_nsxv3/common/synchronization.py
mutax: fixed - please check, reworked the surrounding code.

networking_nsxv3/common/synchronization.py
mutax: what is the advantage/what is your reasoning?

I am not sure if we could always guarantee to clean up properly with the suggested change, the way
I am doing it seems more straight forward and 'saver' -- if I don't miss your point here.

If we first update the counters (which is happening in the done call) and then check if the list is empty, we can always be sure that at that point the list entry gets cleaned up. If we first check if its empty and then update the bookkeeping, we would keep the list around even when it was the only job running.

sven-rosenzweig: I tried to separate the case if ``joblist.empty() == true`` (so no more jobs to execute) and try to return early in this case. From my point of view this would be simpler to understand.

sven-rosenzweig: my comment is meant as optional, so no need to change anything here.

networking_nsxv3/common/synchronization.py
sven-rosenzweig: I guess so.

felix-kaestner (approved):

sven-rosenzweig (approved):

PR #148, opened 2025-03-20
Use logging placeholders for nsx client error msgs

sebageek:
The generic "an error happened" error logging call inside the nsx client now uses the logging placeholders %s instead of calling format() on the log message. This will hopefully fix sentry aggregation, as the initial log message will be the same for all different issues.

I'm also placing my confidence in this patch by fixing the "Argumetns" typo that I previously used to aggregate these errors.


mutax (approved):
lgtm!!!

sven-rosenzweig (approved):

PR #155, opened 2025-06-26
[alternative] Fix duplicate segment port creation due to Nova/vSphere/NSX-T race

sven-rosenzweig:
Fix duplicate segment port creation due to Nova/vSphere/NSX-T race
    
A race condition between Nova, vSphere, and the NSX-T agent can result in
vSphere preemptively creating a segment port before the NSX-T agent does.
These pre-created ports lack security group tags, yet share the same
attachment ID as NSX-T agent managed ports, leading to duplication issues.
    
Key differences in vSphere-created ports:
  - Created by user "system"
  - ID starts with "default:<random_uuid>"
  - ID is not the OpenStack port id
  - Original name format: <random_uuid>vmx@<openstack-port-id>
    
During the NSX-T sync loop, once metadata is updated, we identify these
duplicate ports by matching the attachment ID (which is derived from the
OpenStack port ID). The NSX-T agent does not handle duplicates. So it is
unclear which segment port gets updates.
    
To resolve this:
Before creating a new segment port, check for existing ports with the
same attachment ID and reuse them if found.

### DIFF
--- a/networking_nsxv3/exceptions/agent.py
+++ b/networking_nsxv3/exceptions/agent.py
@@ -1,0 +1,10 @@
+
+from neutron_lib._i18n import _
+from neutron_lib import exceptions
+
+class MultipleSegmentPorts(exceptions.NeutronException):
+    message = _("More thant 2 segment ports found for OpenStack Port %(port_id)s")
+
+class NoneUniqueObjectFound(exceptions.NeutronException):
+    message = _("Found multiple objects for query %(query) found for %(objects)s. "
+                "Please use a more specific query to return a single object.")
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
@@ -6,6 +6,7 @@ from requests.exceptions import ConnectionError, ConnectTimeout, HTTPError
 from oslo_utils import versionutils
 from oslo_log import log as logging
 from oslo_config import cfg
+from networking_nsxv3.exceptions.agent import NoneUniqueObjectFound
 from networking_nsxv3.common.synchronization import Scheduler
 from networking_nsxv3.common.locking import LockManager
 import requests
@@ -228,13 +229,10 @@ class Client(metaclass=Singleton):
     def get_unique(self, path: str, params: dict = dict()) -> dict:
         results = self.get(path=path, params=params).json().get("results")
         if isinstance(results, list):
-            if results:
-                if len(results) > 1:
-                    LOG.error("Ambiguous. %s", results)
-                result = results.pop()
-                return result
-        elif results:
-            return results
+            if len(results) > 1:
+                raise NoneUniqueObjectFound(query=params, objects=results)
+            results = results.pop()
+        return results
 
     def search(self, path: str, params: dict) -> dict:
         res = self.get(path=path, params=params)
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
@@ -236,6 +236,13 @@ class Client(metaclass=Singleton):
         elif results:
             return results
 
+    def search(self, path: str, params: dict) -> dict:
+        res = self.get(path=path, params=params)
+        if not res.ok:
+            #ToDo handle errors properly
+            pass
+        return res.json().get("results")
+
     def get_all(self, path: str, params: dict = None, cursor: str = ""):
         # FYI - NSX does not allow to filter by custom property
         # Search API has hard limit of 50k objects (with cursor)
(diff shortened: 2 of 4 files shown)


networking_nsxv3/exceptions/agent.py
```diff
@@ -0,0 +1,10 @@
+
+from neutron_lib._i18n import _
+from neutron_lib import exceptions
+
+class MultipleSegmentPorts(exceptions.NeutronException):
```
sebageek, line 5: Would call it MultipleSegmentPortsFound, like, with a verb at the end of what action has triggered this.

networking_nsxv3/exceptions/agent.py
```diff
@@ -0,0 +1,10 @@
...
+    message = _("More thant 2 segment ports found for OpenStack Port %(port_id)s")
```
sebageek, line 6: Isn't this for >= 2 aka "more than one" or "2 or more"?

networking_nsxv3/exceptions/agent.py
```diff
@@ -0,0 +1,10 @@
...
+
+class NoneUniqueObjectFound(exceptions.NeutronException):
```
sebageek, line 8: I don't think it's `None` here, but `Non`, like `NonUniqueObjectFound`.

networking_nsxv3/exceptions/agent.py
```diff
@@ -0,0 +1,10 @@
...
+    message = _("Found multiple objects for query %(query) found for %(objects)s. "
```
sebageek, line 9: Two times found used in the message.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -236,6 +236,13 @@ def get_unique(self, path: str, params: dict = dict()) -> dict:
         elif results:
             return results
 
+    def search(self, path: str, params: dict) -> dict:
+        res = self.get(path=path, params=params)
+        if not res.ok:
+            #ToDo handle errors properly
+            pass
```
sebageek, line 241: Should we at least raise an exception or what should happen then?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -967,12 +987,17 @@ def port_realize(self, os_port: dict, delete=False):
...
             os_port["security_groups"] = None
         return self._realize(Provider.PORT, False, self.payload.segment_port, os_port, provider_port)
+
     def get_port(self, os_id):
         port = self.client.get_unique(path=API.SEARCH_QUERY, params={"query": API.SEARCH_Q_SEG_PORT.format(os_id)})
         if port:
             return self.metadata_update(Provider.PORT, port), port
         return None, None
 
+    def find_duplicate_ports(self, os_port_id):
```
sebageek, line 997: Does this really find duplicates? I think this rather `find_ports_os_port()` and then the user checks if multiple ports are there.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -938,7 +940,25 @@ def port_realize(self, os_port: dict, delete=False):
             provider_port["path"] = port_meta.path
             provider_port["_revision"] = port_meta.revision
         else:
-            LOG.info("Segment Port %s not found, creating...", port_id)
+            LOG.info("Segment Port %s not found, creating ...", port_id)
+            # Workaround Race Condition between NSX-T agent and vSphere
+            # In some cases the vSphere creates the Segment Port before NSX-T agent
+            # NSX-T agent will reuse the existing port, otherwise it will create a new one (default behaviour).
+            duplicate_ports = self.find_duplicate_ports(port_id)
+            if duplicate_ports and len(duplicate_ports) > 0:
```
sebageek, line 948: `if duplicate_ports:` should do the same, as `None` and `[]` both evaluate to false.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -938,7 +940,25 @@ def port_realize(self, os_port: dict, delete=False):
             provider_port["path"] = port_meta.path
             provider_port["_revision"] = port_meta.revision
         else:
-            LOG.info("Segment Port %s not found, creating...", port_id)
+            LOG.info("Segment Port %s not found, creating ...", port_id)
+            # Workaround Race Condition between NSX-T agent and vSphere
+            # In some cases the vSphere creates the Segment Port before NSX-T agent
+            # NSX-T agent will reuse the existing port, otherwise it will create a new one (default behaviour).
+            duplicate_ports = self.find_duplicate_ports(port_id)
```
sebageek, line 947: We don't know yet if we have duplicate ports. Should we maybe call this segment ports?

sebageek (commented):
Are there cases where we have to now handle the new id `default:$port_id` or whatever vSphere creates? I know we talked about our internal tooling, but that we can easily fix. Just wanted to make sure you had a look in the driver. :)

sven-rosenzweig:
We will close this PR with respect to https://github.com/sapcc/networking-nsx-t/pull/156

mutax:
in #156 we implemented a fix in cooperation with nova

PR #156, opened 2025-07-02
Fix duplicate segment port creation due to Nova/vSphere/NSX-T race

sven-rosenzweig:
A race condition between Nova, vSphere, and the NSX-T agent can result in
vSphere preemptively creating a segment port before the NSX-T agent does.
These pre-created ports lack security group tags, yet share the same
attachment ID as NSX-T agent managed ports, leading to duplication issues.

Key differences in vSphere-created ports:
 - Created by user "system"
 - ID starts with "default:<random_uuid>"
 - ID is not the OpenStack port id
 - Original name format: <random_uuid>vmx@<openstack-port-id>

During the NSX-T sync loop, once metadata is updated, we identify these
duplicate ports by matching the attachment ID (which is derived from the
OpenStack port ID). The NSX-T agent does not handle duplicates. So it is
unclear which segment port gets updates.

To resolve this:
Send a custom notification that Nova is waiting for to continue the
server creation process. This ensures that the NSX-T agent creates the
segment port before Nova proceeds.

### DIFF
--- a/networking_nsxv3/api/rpc.py
+++ b/networking_nsxv3/api/rpc.py
@@ -121,6 +121,12 @@ class NSXv3AgentRpcClient(object):
         else:
             LOG.debug("NSXv3AgentRpcClient: (no rpc call triggered): pass security_group_id or port_id as tupe ")
 
+    def is_port_realized(self, host, id):
+        """Get the realization status of a resource."""
+        LOG.debug("NSXv3AgentRpcClient: (is_port_realized): %s", id)
+        return self._get_call_context(host).call(
+            self.context, 'is_port_realized', port_id=id)
+
     def update_address_group(self, plugin: Ml2Plugin, payload: events.DBEventPayload):
         address_group = payload.states[1]
         self._get_call_context().cast(self.context, 'address_group_updated', address_group=address_group)
--- a/networking_nsxv3/api/rpc.py
+++ b/networking_nsxv3/api/rpc.py
@@ -1,4 +1,6 @@
 import oslo_messaging
+import neutron.notifiers.nova as nova_notifier
+
 from networking_nsxv3.common import constants as nsxv3_constants
 from networking_nsxv3.db import db
 from neutron_lib import context as neutron_context
@@ -238,6 +240,11 @@ class NSXv3ServerRpcApi(object):
         return cctxt.call(self.context, 'get_address_group_revision_number',
                           address_group_id=address_group_id)
 
+    @log_helpers.log_method_call
+    def send_nova_event(self, event):
+        cctxt = self.client.prepare()
+        return cctxt.call(self.context, 'send_nova_event', event=event)
+
 
 class NSXv3ServerRpcCallback(object):
     """Plugin-side RPC (implementation) for agent-to-plugin interaction.
@@ -362,3 +369,14 @@ class NSXv3ServerRpcCallback(object):
     @log_helpers.log_method_call
     def has_security_group_logging(self, context, security_group_id):
         return db.has_security_group_logging(context, security_group_id)
+
+    @log_helpers.log_method_call
+    def send_nova_event(self, context, event):
+        """Send a custom event to nova."""
+        notifier = nova_notifier.Notifier.get_instance()
+        if not notifier:
+            LOG.error("Nova notifier is not available.")
+            return
+        LOG.info("Sent custom Nova event: %s", event)
+        notifier.send_custom_port_status(event)
+        return
--- a/networking_nsxv3/extensions/nsxtpolicy.py
+++ b/networking_nsxv3/extensions/nsxtpolicy.py
@@ -1,0 +1,125 @@
+import abc
+import json
+import importlib
+import functools
+
+#Load common config here, as registering the api extension fails without
+from neutron.common import config
+config.register_common_config_options()
+
+from neutron.api import extensions
+from neutron.api.v2.resource import Resource
+from neutron_lib.api import extensions as api_extensions
+
+from neutron_lib.api import faults
+from neutron import policy
+
+try:
+    from neutron.api import wsgi
+except ImportError:
+    from neutron import wsgi
+
+from webob import exc as web_exc
+from webob import exc as exceptions
+from oslo_log import log
+
+import networking_nsxv3.extensions
+
+LOG = log.getLogger(__name__)
+
+ACCESS_RULE = "context_is_cloud_admin"
+
+def check_cloud_admin(f):
+    @functools.wraps(f)
+    def wrapper(self, request, *args, **kwargs):
+        if not policy.check(request.context, ACCESS_RULE, {'project_id': request.context.project_id}):
+            raise web_exc.HTTPUnauthorized("{} required for access".format(ACCESS_RULE))
+        return f(self, request, *args, **kwargs)
+    return wrapper
+
+class NSXTAPIDefinition:
+    NAME = "NSXT Policy API"
+    ALIAS = "nsxt-policy"
+    DESCRIPTION = "Expose NSXT Policy API via Neutron API"
+    UPDATED_TIMESTAMP = "2025-07-25T09:25:00+02:00"
+    RESOURCE_ATTRIBUTE_MAP = {}
+    SUB_RESOURCE_ATTRIBUTE_MAP = {}
+    REQUIRED_EXTENSIONS = []
+    OPTIONAL_EXTENSIONS = []
+
+
+class Nsxtpolicy(api_extensions.APIExtensionDescriptor):
+    """NSX policy API extensions"""
+    # class name cannot be camelcase, needs to be just capitalized
+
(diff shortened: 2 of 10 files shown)


networking_nsxv3/api/rpc.py
```diff
@@ -119,6 +121,12 @@ def trigger_manual_update(self, id, type):
         else:
             LOG.debug("NSXv3AgentRpcClient: (no rpc call triggered): pass security_group_id or port_id as tupe ")
 
+    def get_realization_status(self, host, id):
+        """Get the realization status of a resource."""
+        LOG.debug("NSXv3AgentRpcClient: (get_realization_status): " + str(id))
```
mutax: `LOG.debug("NSXv3AgentRpcClient: (get_realization_status): %s", id)`

networking_nsxv3/api/rpc.py
```diff
@@ -238,6 +246,13 @@ def get_address_group_revision_number(self, address_group_id):
         return cctxt.call(self.context, 'get_address_group_revision_number',
                           address_group_id=address_group_id)
 
+    @log_helpers.log_method_call
+    def send_nova_event(self, event):
+        LOG.info("RPC notify nova event %s", event)
+        cctxt = self.client.prepare()
+        LOG.info("RPC notify 2 nova event %s", event)
```
mutax: do we need two info logs here? 
would one debug log be sufficient?

networking_nsxv3/api/rpc.py
```diff
@@ -238,6 +246,13 @@ def get_address_group_revision_number(self, address_group_id):
...
+        return cctxt.call(self.context, 'send_nova_event',event=event)
```
mutax: pep - space missing between `,event`

networking_nsxv3/api/rpc.py
```diff
@@ -362,3 +377,15 @@ def get_address_group_revision_number(self, context, address_group_id):
     @log_helpers.log_method_call
     def has_security_group_logging(self, context, security_group_id):
         return db.has_security_group_logging(context, security_group_id)
+
+    @log_helpers.log_method_call
+    def send_nova_event(self, context, event):
+        """Send a custom event to nova."""
+        LOG.info("RPC received notify nova event %s", event)
```
mutax: there is already a log_method_call -- do we need another log with info here?

networking_nsxv3/extensions/nsxtpolicy.py
```diff
@@ -0,0 +1,126 @@
...
+        if not policy.check(request.context, ACCESS_RULE, {'project_id': request.context.project_id}):
+            raise web_exc.HTTPUnauthorized("{} required for access".format(ACCESS_RULE))
+        return f(self, request, *args, **kwargs)
+    return wrapper
+
+class NSXTAPIDefinition:
+    NAME = "NSXT Policy API"
+    ALIAS = "nsxt-policy"
+    DESCRIPTION = "Expose NSXT Policy API via Neutron API"
+    UPDATED_TIMESTAMP = "2021-08-25T18:18:42+02:00"
```
mutax: update the timestamp?

networking_nsxv3/extensions/nsxtpolicy.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @check_cloud_admin
+    def show(self, request, **kwargs):
+        port_id = request.params.get("port_id")
+        binding_host = request.params.get('binding_host')
+
+        if port_id:
+            realization_status = self._port_realization_status(port_id, binding_host)
+        else:
+            LOG.error("No port_id or binding_host provided in request params")
```
mutax: the if only checks port_id not binding_host. 
`if port_id and binding_host:`

networking_nsxv3/extensions/nsxtpolicy.py
```diff
@@ -0,0 +1,126 @@
...
+            raise web_exc.HTTPBadRequest("Either port_id and binding_host must be provided")
```
mutax: "Both" not Either?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py
```diff
@@ -133,10 +135,18 @@ def delete_log_precommit(self, context, log_obj):
...
         pass
-    
+
     def address_group_updated(self, context, address_group):
         self.callback(address_group, self.realizer.address_group_update)
 
+    def get_realization_status(self, context, port_id):
+        try:
+            realization_status = self.realizer.port_realization_status(port_id)
+        except Exception:
```
mutax: can we log that exception as error? otherwise it will be silently dropped and this will bite us when debugging

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
```diff
@@ -284,6 +285,41 @@ def port(self, os_id: str):
                 LOG.info("deletion realization of port %s", os_id)
                 self._port_realize({"id": os_id}, delete=True)
 
+    def port_realization_status(self, port_id):
+        LOG.info("Fetching the port realization status for port_id: %s", port_id)
+
+        res = self.nsx_provider.client.get(path=API.SEARCH_QUERY, params={"query": API.SEARCH_Q_SEG_PORT.format(port_id)})
+        results = res.json()
+
+        segment_port = None
```
mutax: proposing plural, segment_ports, as this will be a list

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
```diff
@@ -284,6 +285,41 @@ def port(self, os_id: str):
...
+
+        if not segment_port:
+            LOG.error("No segment port found for port_id: %s", port_id)
+            return False
+
+        if len(segment_port) > 1:
+            LOG.error("Multiple segment port found for port_id: %s.", port_id)
+            return False
+
+        segment_port = segment_port[0]
```
mutax: see above.` segment_port = segment_ports[0]` is easier to comprehend

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py
```diff
@@ -1,4 +1,5 @@
 import eventlet
+import neutron.notifiers.nova as nova
```
mutax: maybe would call it nova_notifier - from the name I would expect this to be a nova client when just reading the code

networking_nsxv3/api/rpc.py
```diff
@@ -1,4 +1,6 @@
 import oslo_messaging
+import neutron.notifiers.nova as nova
```
mutax: as below, would call the import nova_notifier or similar, just from the name I would expect this to be a nova client when I look at the code

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/agent.py
```diff
@@ -133,10 +135,18 @@ def delete_log_precommit(self, context, log_obj):
 
     def resource_update(self, context, log_obj):
         pass
-    
+
     def address_group_updated(self, context, address_group):
         self.callback(address_group, self.realizer.address_group_update)
 
+    def get_realization_status(self, context, port_id):
```
toanju:
```suggestion
    def is_port_realized(self, context, port_id):
```

otherwise I'd kind of expect something else than a boolean to be returned (same for the callers)

networking_nsxv3/api/rpc.py
sven-rosenzweig: No, used this log for debugging purposes. As I had issues in the agent to driver communication

sven-rosenzweig: Removing both.

networking_nsxv3/api/rpc.py
sven-rosenzweig: Agree with you we do not need this. Used this log for debugging purposes.

networking_nsxv3/tests/unit/openstack.py
```diff
@@ -384,3 +384,6 @@ def get_qos(self, os_id):
     def has_security_group_logging(self, security_group_id):
         g = self.inventory.get_by_id(NeutronMock.SECURITY_GROUP, security_group_id)
         return g is not None and g.get("logged")
+
+    def send_nova_event(self, event):
```
mutax, line 388: in the real code calling this method returns the return value of the nova call - would have an implicit return here with a value?

sven-rosenzweig: We put our custom nova event on a notifier queue, the notifier is processing this (also handling errors). So we actually do not care about the response.

mutax (approved):
lgtm

rhalina (approved):

PR #157, opened 2025-07-21
 Adjust sentry fingerprinting and Add metrics for API call durations and errors

sven-rosenzweig:
Adjust fingerprinting to reduce event noise for NSX-T API calls
    
Sentry groups events by fingerprint. This change customizes the fingerprinting
for NSX-T-related errors to avoid generating separate events for each API call
with a unique ID. The replace operation does not work in-place, so the adjustment
ensures consistent grouping.

-----

Add metrics for API call durations and errors
    
This commit introduces Prometheus metrics to capture details about outgoing
API calls, including their duration and status codes. Additionally, a separate
to track request-level errors such as connection timeouts and network failures.
These metrics provide better observability into API behavior and help identify
performance or reliability issues.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
@@ -44,12 +44,13 @@ class Singleton(type):
 
 class RetryPolicy(object):
 
-    def _create_sentry_fingerprint(path: str):
-        # check if uuid is part of path -> replace with {}
+    @staticmethod
+    def _create_sentry_fingerprint(path: str, placeholder: str = "{}") -> str:
+        # check if uuid is part of path -> replace with placeholder
         for sub in path.split("/"):
             try:
                 uuid.UUID(sub)
-                path.replace(sub, "{}")
+                path = path.replace(sub, placeholder)
             except ValueError:
                 pass
         return path
@@ -121,6 +122,9 @@ class RetryPolicy(object):
                 LOG.debug(msg)
                 eventlet.sleep(pause)
 
+            m = response.request.method if response else "UNKNOWN"
+            sentry_extra["fingerprint"] = [RetryPolicy._create_sentry_fingerprint(kwargs.get("path", '')), m]
+            LOG.exception(last_err, extra=sentry_extra)
             raise RuntimeError(msg, last_err)
 
         return decorator
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
@@ -3,7 +3,7 @@ import eventlet
 eventlet.monkey_patch()
 
 from requests import Response
-from requests.exceptions import ConnectionError, ConnectTimeout, HTTPError
+from requests.exceptions import ConnectionError, ConnectTimeout, HTTPError, ReadTimeout
 from oslo_utils import versionutils
 from oslo_log import log as logging
 from oslo_config import cfg
@@ -151,7 +151,7 @@ class RetryPolicy(object):
                                                        response.request.method]
                         LOG.error("Request=%s Response=%s", request_info, last_err, extra=sentry_extra)
                         break
-                except (HTTPError, ConnectionError, ConnectTimeout) as err:
+                except (HTTPError, ConnectionError, ConnectTimeout, ReadTimeout) as err:
                     last_err = err
                     m = response.request.method if response else "UNKNOWN"
                     sentry_extra["fingerprint"] = [RetryPolicy._create_sentry_fingerprint(kwargs.get("path", '')), m]
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
@@ -1,4 +1,5 @@
 import eventlet
+
 eventlet.monkey_patch()
 
 from requests import Response
@@ -8,6 +9,7 @@ from oslo_log import log as logging
 from oslo_config import cfg
 from networking_nsxv3.common.synchronization import Scheduler
 from networking_nsxv3.common.locking import LockManager
+from networking_nsxv3.prometheus import exporter
 import requests
 import uuid
 import time
@@ -43,9 +45,26 @@ class Singleton(type):
 
 
 class RetryPolicy(object):
+    BASE = "/policy/api/v1"
+    BASE_INFRA = f"{BASE}/infra"
+    RULES = [
+        # Regex Order matters
+        # Match security_rule before security_policy
+        # Match ports before segments
+        ('security_rule', re.compile(f"^{BASE_INFRA}/domains/default/security-policies/.*/rules.*")),
+        ('security_policy', re.compile(f"^{BASE_INFRA}/domains/default/security-policies.*")),
+        ('group', re.compile(f"^{BASE_INFRA}/domains/default/groups.*")),
+        ('port', re.compile(f"^{BASE_INFRA}/segments/.*/ports.*")),
+        ('segments', re.compile(f"^{BASE_INFRA}/segments.*")),
+        ('realized_state', re.compile(f'^{BASE_INFRA}/realized-state/status.*')),
+        ('services', re.compile(f"^{BASE_INFRA}/services.*")),
+        ('search', re.compile(f"^{BASE}/search.*")),
+        ('transport_zone', re.compile(f"^{BASE_INFRA}/sites/default/enforcement-points/default/transport-zones.*")),
+        ('qos_profile', re.compile(f"^{BASE_INFRA}/default/qos-profiles.*")),
+    ]
 
-    @staticmethod
-    def _create_sentry_fingerprint(path: str, placeholder: str = "{}") -> str:
+    @classmethod
+    def _create_sentry_fingerprint(cls, path: str, placeholder: str = "{}") -> str:
         # check if uuid is part of path -> replace with placeholder
         for sub in path.split("/"):
             try:
@@ -54,6 +73,25 @@ class RetryPolicy(object):
             except ValueError:
                 pass
         return path
+
+    @classmethod
+    def _get_resource_type(cls, path: str) -> str:
+        for name, rule in cls.RULES:
+            match = rule.match(path)
+            if match:
+                return name
+        return "unknown"
+
+    @classmethod
+    def _update_metric(cls, metric, path='', status='UNKNOWN', method='UNKNOWN', response_time=0, exception_type='UNKNOWN'):
+        fp_path = cls._create_sentry_fingerprint(path=path, placeholder="<uuid>")
+        resource = cls._get_resource_type(path)
+
+        if exporter.API_CALLS == metric:
+            metric.labels(method=method, bb=cfg.CONF.host, resource_type=resource, path=fp_path, status=status).observe(response_time)
+
+        if exporter.API_CALL_EXCEPTIONS == metric:
+            metric.labels(bb=cfg.CONF.host, resource=resource, path=path, exception_type=exception_type).inc()
 
     def __call__(self, func):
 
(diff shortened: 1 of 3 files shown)


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -44,16 +47,66 @@ def __call__(cls, *args, **kwargs):
...
             try:
                 uuid.UUID(sub)
-                path.replace(sub, "{}")
+                path =  path.replace(sub, placeholder)
             except ValueError:
                 pass
         return path
 
+    @staticmethod
+    def get_resource_type(path: str) -> str:
```
mutax: I think adding a comment with all the expected paths and what we return would be helpful here, and explain why the matching is a bit more complex. One example path before every return maybe?

mutax: alternatively we could come up with a regex for each instance, and store a compiled version in a list and iterate over them, but that would require being careful with the order of the entries to prevent partial matches. 
```python
import re
path = "domains/default/security-policies/"

rules = [
  ('security_rule', re.compile('^domains/default/security-policies/.*security_rule.*')),
  ('security_policy', re.compile('^domains/default/security-policies/.*')),
  ]

for name, rule in rules:
    match = rule.match(path)
    if match:
        print(match)
        print(name)
```
but I am not sure if that makes it better to understand?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -44,16 +47,66 @@ def __call__(cls, *args, **kwargs):
...
+        if path.startswith("realized-state/status"):
+            return "realization_status"
+
+        return "unknown"
+
+    @staticmethod
+    def _update_metric(metric, **kwargs):
+        path = kwargs.get("path", '')
+        new_path = RetryPolicy._create_sentry_fingerprint(path=path, placeholder=":id")
+        resource = RetryPolicy.get_resource_type(path)
```
mutax: I would make the naming consistent, either both helper methods should start with a _ or none :)

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -44,16 +47,66 @@ def __call__(cls, *args, **kwargs):
...
+
+        if path.startswith("realized-state/status"):
+            return "realization_status"
+
+        return "unknown"
+
+    @staticmethod
+    def _update_metric(metric, **kwargs):
+        path = kwargs.get("path", '')
+        new_path = RetryPolicy._create_sentry_fingerprint(path=path, placeholder=":id")
```
mutax: is the fingerprint shown to the user? is `:id` something used elsewhere? `<uuid>` feels more straight forward to me (huge nitpick)

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -44,16 +47,66 @@ def __call__(cls, *args, **kwargs):
...
+        resource = RetryPolicy.get_resource_type(path)
+
+        if isinstance(metric, prometheus_client.Histogram):
```
mutax: I would rather match on the name of the metric instead of the class - this implementation makes it hard to add more metrics later and when reading the code I have no idea what the resulting metric is called. 

But I also don't see why we need to have one method to handle all metrics and not just add one update method per metric, then no check is required.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -107,20 +162,26 @@ def decorator(self, *args, **kwargs):
                     if not is_atomic_request_error(response):
                         # skip retry on the ramaining NSX errors
                         sentry_extra["fingerprint"] = [RetryPolicy._create_sentry_fingerprint(kwargs.get("path", '')),
-                                                       response.request.method]
+                                                       response.request.method,
+                                                       "sven test 1"]
```
mutax: "sven test 1" ?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,59 @@ def __call__(cls, *args, **kwargs):
...
+    ]
+
+    @staticmethod
+    def _create_sentry_fingerprint(path: str, placeholder: str = "{}") -> str:
         # check if uuid is part of path -> replace with {}
         for sub in path.split("/"):
             try:
                 uuid.UUID(sub)
-                path.replace(sub, "{}")
+                path =  path.replace(sub, placeholder)
```
mutax: pep: 2 spaces after =

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,59 @@ def __call__(cls, *args, **kwargs):
 
 
 class RetryPolicy(object):
-
-    def _create_sentry_fingerprint(path: str):
+    BASE = "/policy/api/v1"
+    BASE_INFRA = "/policy/api/v1/infra"
```
mutax: could replace it with `BASE_INFRA = f"{BASE}/infra"` so one can view immediately that they are related?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,59 @@ def __call__(cls, *args, **kwargs):
...
                 uuid.UUID(sub)
-                path.replace(sub, "{}")
+                path =  path.replace(sub, placeholder)
             except ValueError:
                 pass
         return path
 
+    @staticmethod
+    def _get_resource_type(path: str) -> str:
+        for name, rule in RetryPolicy.RULES:
```
mutax: Not sure if a classmethod would make more sense, then `RetryPolicy.RULLES` could be rewritten as `cls.RULES` and would not be hardcoded, and in case at some point this class is subclassed this will not reference the wrong rules? Possible nitpick.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,59 @@ def __call__(cls, *args, **kwargs):
...
+            match = rule.match(path)
+            if match:
+                return name
+        return "unknown"
+
+    @staticmethod
+    def _update_metric(metric, **kwargs):
+        path = kwargs.get("path", '')
+        new_path = RetryPolicy._create_sentry_fingerprint(path=path, placeholder="<uuid>")
```
mutax: same as with line 77 - maybe classmethod?

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -76,7 +120,9 @@ def decorator(self, *args, **kwargs):
                 try:
                     response = func(self, *args, **kwargs)
                     # LOG.debug("REQUEST: %s STATUS: %s, RESPONSE.CONTENT %s", requestInfo, response.status_code, response.content)
-
+                    RetryPolicy._update_metric(exporter.API_CALLS, path=kwargs.get("path", ''), method=response.request.method,
```
mutax: hm, I see the problem, maybe a retry_policy = self at the start of the __call__ method and then reference that one? They way RetryPolicy is used (all with staticmethods) you could also put the methods in the global namespace of this file instead and reference them from there - no need to put it into the class then.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
sven-rosenzweig: I like that.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
sven-rosenzweig: As discussed in Slack, we stick with the current staticmethod implementation

mutax (dismissed):

networking_nsxv3/tests/unit/test_retry_policy.py
```diff
@@ -0,0 +1,23 @@
...
+            ('security_policy', f"/policy/api/v1/infra/domains/default/security-policies"),
+            ('security_policy', f"/policy/api/v1/infra/domains/default/security-policies/{os_id}"),
+            ('group', f"/policy/api/v1/infra/domains/default/groups/{os_id}"),
+            ('port', f"/policy/api/v1/infra/segments/{os_id}/ports"),
+            ('segments', f"/policy/api/v1/infra/segments/{os_id}")
+        ]
+
+        for resource_type, path in paths:
+            classified = RetryPolicy._get_resource_type(path)
+            print(resource_type, classified)
```
sebageek: `print()` could be removed

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,59 @@ def __call__(cls, *args, **kwargs):
...
+                return name
+        return "unknown"
+
+    @staticmethod
+    def _update_metric(metric, **kwargs):
+        path = kwargs.get("path", '')
+        new_path = RetryPolicy._create_sentry_fingerprint(path=path, placeholder="<uuid>")
+        resource = RetryPolicy._get_resource_type(path)
+
+        if "histogram:nsxv3_agent_api_calls" == str(metric):
```
sebageek: You could use `if str(RetryPolicy.API_CALLS) == str(metrics):` or probably even without the `str()`

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
sebageek: What was the result of that discussion? classmethod sounds good ;)

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,59 @@ def __call__(cls, *args, **kwargs):
...
+    def _get_resource_type(path: str) -> str:
+        for name, rule in RetryPolicy.RULES:
+            match = rule.match(path)
+            if match:
+                return name
+        return "unknown"
+
+    @staticmethod
+    def _update_metric(metric, **kwargs):
+        path = kwargs.get("path", '')
```
sebageek: You have a lot of `kwargs.get()` with default arguments. How about removing the `**kwargs` and add them all explicitly with a default? It doesn't really look like you need the variability unspecific arguments.

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,54 @@ def __call__(cls, *args, **kwargs):
...
                 uuid.UUID(sub)
-                path.replace(sub, "{}")
+                path = path.replace(sub, placeholder)
             except ValueError:
                 pass
         return path
 
+    @classmethod
+    def _get_resource_type(cls, path: str) -> str:
+        for name, rule in RetryPolicy.RULES:
```
sebageek: I think this `RetryPolicy` should also be transformed to `cls`  now that we're using `@classmethod`

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
```diff
@@ -43,17 +45,54 @@ def __call__(cls, *args, **kwargs):
...
+    def _get_resource_type(cls, path: str) -> str:
+        for name, rule in cls.RULES:
+            match = rule.match(path)
+            if match:
+                return name
+        return "unknown"
+
+    @classmethod
+    def _update_metric(cls, metric, path='', status='UNKNOWN', method='UNKNOWN', response_time=0, exception_type='UNKNOWN'):
+        new_path = cls._create_sentry_fingerprint(path=path, placeholder="<uuid>")
```
sebageek: [optional] Could name this `fp_path` or something to signify that this is for the fingerprint.

sebageek (dismissed):

networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/client_nsx.py
sven-rosenzweig: done

sebageek (approved):

toanju (approved):

PR #158, opened 2025-07-30
fix: Nova gets wrong port realization status

sven-rosenzweig:
In some cases, Nova does not wait for the network-realized custom event, as this may be sent too early. Therefore, an API extension was developed to allow checking the status of a segment port explicitly when the event is not available.

The check for whether a port is realized or not should match the logic used in the custom event sent once realization is complete. Thus, we remove the publish_state.

### DIFF
--- a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
+++ b/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
@@ -288,31 +288,29 @@ class AgentRealizer(object):
     def port_realization_status(self, port_id):
         LOG.info("Fetching the port realization status for port_id: %s", port_id)
 
-        res = self.nsx_provider.client.get(path=API.SEARCH_QUERY, params={"query": API.SEARCH_Q_SEG_PORT.format(port_id)})
-        results = res.json()
+        port_meta = self.nsx_provider.metadata(self.nsx_provider.PORT, port_id)
+        if not port_meta:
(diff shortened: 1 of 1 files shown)


networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/realization.py
```diff
@@ -310,9 +310,11 @@ def port_realization_status(self, port_id):
 
         segment_port = segment_ports[0]
         status = segment_port.get("status")
-        if status["publish_status"] == "REALIZED" and status["consolidated_status"]["consolidated_status"] == "SUCCESS":
+        if status["consolidated_status"]["consolidated_status"] == "SUCCESS":
```
mutax: imho this is more likely to give a key error compared to the previous version.
there we first checked if `publish_status` is realized. if that check fails the second part after the and will not be checked.
would add a try, log the key error and return false instead.,

joker-at-work: Right, the [other code](https://github.com/sapcc/networking-nsx-t/blob/f261cd8c87fbedb90f311f6abe179a6634cbae5a/networking_nsxv3/plugins/ml2/drivers/nsxv3/agent/provider_nsx_policy.py#L736) expects things to not be there from the way it's written.

mutax (dismissed):

joker-at-work (dismissed):

mutax (approved):

rhalina (approved):

